Issue #35, opened 2017-11-25
<Toolbar> with fixed and flexible does not get inline css margin-top applied to <ToolbarFixedAdjust>


token419:
First off, awesome work, I love this package since i found it last night!

I noticed that when using a <Toolbar> with the "fixed" and "flexible" properties that the <ToolbarFixedAdjust> remains at the default margin-top applied by css, but is not being overridden as it should via an inline style from javascript. I'm guessing the functionality was just over looked.

The expected functionality from the material web components can be seen here: https://material-components-web.appspot.com/toolbar/waterfall-flexible-toolbar.html

If you inspect their source, you can see they have added the following to achieve the effect (I omitted the non relevant parts):
```
function init() {
    var toolbar = mdc.toolbar.MDCToolbar.attachTo(document.querySelector('.mdc-toolbar'));
    toolbar.fixedAdjustElement = document.querySelector('.mdc-toolbar-fixed-adjust');
}
```

Code to reproduce via these components (you will not be able to see the paragraph tag because the margin is only 1 row, and not the default 4 rows from mdc-toolbar--flexible-default-behavior):
```
<Toolbar fixed="True" flexible="True" class="mdc-toolbar--flexible-default-behavior">
    <ToolbarRow>
        <ToolbarSection alignStart>
            <ToolbarMenuIcon use="menu"/>
            <ToolbarTitle>Title</ToolbarTitle>
        </ToolbarSection>
    </ToolbarRow>
</Toolbar>
<main>
    <ToolbarFixedAdjust></ToolbarFixedAdjust>
    <p>This will be unreadable</p>
</main>
```

I was able to resolve this by adding the following missing lines to the "Toolbar" "onUpdate" function:
```
var toolbarquery = _mdc.MDCToolbar.attachTo(document.querySelector('.mdc-toolbar'));
toolbarquery.fixedAdjustElement = document.querySelector('.mdc-toolbar-fixed-adjust');
```
I also verified that the correct behavior takes place when using the CSS override via ID like so:
```
#my-flexible-header {
    --mdc-toolbar-ratio-to-extend-flexible: 8;
}
```
If you add that ID to the toolbar, you will notice, the content now expands the full 8 rows instead of the default 4. I also tested with various other toolbar setups, including with no flags and saw no regression from the change.

I'd be happy to submit a pull request if you would like. As far as i can tell, this does not break any other toolbars, but please QA as necessary.

jamesmfriedman:
[...]

PR #37, opened 2017-11-25
Toolbar Fixed Flexable margin-top fix - issue 35

token419:
Created pull request for #35

### DIFF
--- a/src/Toolbar/index.js
+++ b/src/Toolbar/index.js
@@ -115,6 +115,8 @@ export const Toolbar = withMDC({
   mdcConstructor: MDCToolbar,
   mdcElementRef: true,
   onUpdate(props, nextProps, api, inst) {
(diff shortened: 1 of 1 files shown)


jamesmfriedman:
Thanks for the PR @token419! I'm not going to be able to merge this in, but it did help me easily identify the issue.

- using document.querySelector breaks React's DOM abstraction api. This will prevent server side rendering from working, and is considered bad practice for React unless you have no other options.
- your fix assumes there is only one toolbar and one toolbar fixed adjust element on the screen. If there are multiple, this will fail.

I'm working on a fix right now, I'll let you know when it's up.

token419:
Perfect, yeah no worries at all! I figured a better approach would be to hook in with the componentDidMount and componentWillUnmount so as to properly unload the event listener, but it was late and I'm still learning/playing with react.

I'm going to go ahead and close this one out.

Issue #49, opened 2017-12-08
Callbacks should be optional


j-o-d-o:
Things like onChange, onClick, onSelect etc. should always be optional. E.g. on the menus: https://jamesmfriedman.github.io/rmwc/menus if you open the console and select a menu item, there will be a TypeError in the console that onSelect is not a function. ~~Same issue with the Dialogs onClose()~~. 

It is not breaking anything, just annoying to have a bunch of Errors in the console.
I think all it takes is instead of calling the callback such as: 
`props.onSelected(evt); `
do this: 
`props.onSelected && props.onSelected(evt);`

I can take a look where I think it is missing and submit a pr if you wish.

jamesmfriedman:
That would be excellent. I would love the contribution. 

In a lot of be api, I think I just set the default prop to a noop callback so I never had to think about it. PR would be welcomed.

PR #50, opened 2017-12-08
fix(menu): Add missing noop to onClose() & onSelect() callbacks

j-o-d-o:
The noop defaults where actually only missing on the Select component :). 

Fixes: #49


jamesmfriedman:
You rock @j-o-d-o, Thanks for being the first contributor!

PR #52, opened 2017-12-11
fix(LayoutGrid): export GridInner

j-o-d-o:
Export GridInner to the global export. It's useful when dealing with multiple nested grids. Of course, for the time being, it is always possible to import: `import { GridInner } from 'rmwc/Grid';` as discussed on another occasion.
Maybe let's also add more info to the documentation? I had to look into the code to make sure what `<Grid/>` is doing.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -57,7 +57,7 @@ export {
   GridTileTitleSupportText
 } from './GridList';
 
-export { Grid, GridCell } from './Grid';
+export { Grid, GridCell, GridInner } from './Grid';
 
 export { Icon } from './Icon';
 


j-o-d-o:
So this example:

```
<!-- RMWC -->
<Grid>
  <GridCell>
    <Grid>
      <GridCell>2. level</GridCell>
    </Grid>
  </GridCell>
  <GridCell>1. level</GridCell>
</Grid>

<!-- will produce this MDC: -->
<div class="mdc-layout-grid">
  <div class="mdc-layout-grid__inner">
    <div class="mdc-layout-grid__cell">
      <div class="mdc-layout-grid"> <!-- This shouldn't be here according to MDC documentation -->
        <div class="mdc-layout-grid__inner">
          <div class="mdc-layout-grid__cell">2. level</div>
        </div>
      </div>
    </div>
    <div class="mdc-layout-grid__cell">1. level</div>
  </div>
</div>
```

By using `<GridInner/>` instead of `<Grid/>` when it is nested, the second `<div class="mdc-layout-grid">` will not be generated. Yes, it works just fine as it is right now. Yet, I feel more comfortable staying as close to the original MDC as possible as you never know what changes might come up the road. 

That being said, it is really something minor, nothing to worry about too much.

PR #60, opened 2017-12-15
feat(TextField): Add valid prop

UFOMelkor:
Currently there is no possibility to mark a TextField as invalid while mdc provides one.

Please be aware: I'm new to React.JS, thus I guess my PRs are possibly far away from being perfect and may contain errors. If there are better ways to do, please do not hesitate to notice, correct or close.

### DIFF
--- a/src/TextField/index.js
+++ b/src/TextField/index.js
@@ -108,6 +108,8 @@ type TextFieldPropsT = {
   inputRef?: React.Ref<any>,
   /** Disables the input. */
   disabled?: boolean,
+  /** Sets the Validity state of the input */
+  valid?: boolean,
   /** A label for the input. */
   label?: React.Node,
   /** Add a leading icon. */
@@ -125,6 +127,7 @@ export const TextField = withMDC({
   defaultProps: {
     inputRef: noop,
     disabled: false,
+    valid: undefined,
     box: undefined,
     fullwidth: undefined,
     label: undefined,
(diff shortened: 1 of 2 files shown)


jamesmfriedman:
Hi @UFOMelkor, thanks for the PR. Could you give me a little more explanation?

- I looked through the MDC docs and the closest thing I could find is marking a field as invalid which gives the root of the TextField a className of `mdc-text-field--invalid`
- You added the valid prop to the type and defaultProps, but it doesn't actually do anything right now. This would just simple add a Dom attribute called "valid"
- We can put an issue in to support marking a field as `invalid` which looks like it would be compliant with the MDC spec.

j-o-d-o:
What you can do in mdc is this: 
```
let my_textfield = mdc.textField.MDCTextField.attachTo(document.querySelector('.mdc-text-field'));

if (my_great_and_complicated_condition_fails)
  my_textfield.valid = false;
```
This will basically add `mdc-text-field--invalid` to the textfield div. 

Rmwc exposes the api via the `apiRef={api => this.menuCtrl = api}` attirbute. So you could do this:
```
<TextField apiRef={api => this.api= api} />

if (my_great_and_complicated_condition_fails)
  this.api.valid = false;
```

jamesmfriedman:
I'm going to close out this PR but I opened a ticket to look into this as an enhancement.

PR #62, opened 2017-12-18
MDC 27

jamesmfriedman:
@j-o-d-o @UFOMelkor @Kikketer @ncknt

Made a bunch of updates related to Selects and also getting up to speed with MDC27.

Any community feedback would be greatly appreciated, going to start working on the docs but there are working examples in storybook `npm run storybook`.

### DIFF
--- a/docs/content.js
+++ b/docs/content.js
@@ -486,7 +486,7 @@ export default [
       'List',
       'ListItem',
       'ListItemText',
-      'ListItemTextSecondary',
+      'ListItemSecondaryText',
       'ListItemStartDetail',
       'ListItemEndDetail',
       'ListDivider',
--- a/docs/content.js
+++ b/docs/content.js
@@ -441,14 +441,14 @@ export default [
       'https://material.io/components/web/catalog/input-controls/text-fields/',
     example: `
 				<TextField label="Write something..." />
-				<TextFieldHelpText>Optional help text.</TextFieldHelpText>
+				<TextFieldHelperText>Optional help text.</TextFieldHelperText>
 
 				<TextField withLeadingIcon={<TextFieldIcon use="search"/>} label="Write something..." /><br/>
 
 				<TextField box withTrailingIcon={<TextFieldIcon use="close"/>} label="Write something..." /><br/>
 				
 				<TextField textarea fullwidth label="Multiline..." rows="8" />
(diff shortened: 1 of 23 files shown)


j-o-d-o:
Cool stuff! Thanks for the updates!

When creating a Button with pure MDC, the ripple is light. When creating one with RMWC (and setting primary and accent to false (which should be the default imho)), the ripple is dark. The difference is that in RMWC the 'mdc-ripple-surface' class is added, while in MDC the classes is not added to the button. Not sure how it should be. 
Having control of the ripple color with primary and accent is cool. But I would set the 'primary' attribute to false. At least when the butten is raised or uneleveted because then the ripple is basically not noticable as the background then has the same color as the ripple.

I looked at all the other code, but haven't tested everything into detail. But what I have tested and seen looks good!

jamesmfriedman:
Their doc pages are erroring out right now (apparently they are over quota on their own project). 

I'll double check as soon as they come back online. I thought the button ripple color was aligned with how it was supposed to be in the docs, but I might not have checked close enought.

j-o-d-o:
Yeah, saw that too. Here is a codepen using the latest mdc version: https://codepen.io/j-o-d-o/pen/KmBRMo

jamesmfriedman:
Ah ok I think I see what you're talking about. Thanks for the example.

If you remove the --raised modifier from your button, you'll notice the ripple is primary colored. In RMWC, the ripple was gray, which is why I added the primary and accent modifiers (which I didn't want to do).

So maybe this is a timing issue? Seems like your example automatically applies the correct color ripple without further configuration.

jamesmfriedman:
I still hate that I haven't gotten around to doing some proper docs, but at least the Select component has some documentation around it, regardless of how ugly...

New docs for the new year.

j-o-d-o:
@jamesmfriedman great stuff! Thanks man! 
And regarding the documentation. I find it actually quite intuitive and helpful. I don't have to read through never-ending blocks of text, but just have some nice examples and know what kind of API I can use. 
The only thing that is documented, but might deserve a dedicated page with examples is the "Some API wide features" section on the getting started page. It is easy to miss and not super clear on how to use them. 

Thou my opinion might be a little biased as I know mdc quite well and also the rmwc code to a certain extent.

PR #64, opened 2017-12-21
FIx console.log in Select.

izakfilmalter:


jamesmfriedman:
Thanks for catching this, its not actually meant to even be in there anymore, was just debugging, but I'll merge this in and remove it later.

izakfilmalter:
Ya I keep getting an error so idk.

jamesmfriedman:
I'm pretty sure it's related to a timing issue with MDC and the virtual DOM. it's hard to know when to call resize, so it's easier to just call it everytime and catch the error. As soon as I stop logging it, you won't see it anymore.

izakfilmalter:
Ya I'm fairly sure this triggers when the view gets destroyed and it trys
to update it self.

PR #69, opened 2017-12-25
feat(TabBarScroller): Adding TabBarScroller component.

vdkdev:
Using MDCTabBarScroller from @material/tabs/dist/mdc.tabs.


jamesmfriedman:
Thanks for your work on this @vdkdev. There are a couple of edge cases I want to look into, but this seems to be mostly there and working. Going to merge it in for now and it will definitely be going into the next release.

PR #71, opened 2017-12-27
fix(TabBar): reinit when adding/removing tabs dynamically.

### LINKED ISSUE #67: Tabs added after initial render are not upgraded
### What RMWC Version are you using?

0.0.1-rc12

### What browser(s) and React Version is this bug affecting?

Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/63.0.3239.84 Safari/537.36

React ^16.2.0

### What build system are you using?

Webpack

### What are the steps to reproduce the bug?

Create an app where the number of tabs dynamically changes from 1 to 2.

### What is the expected behavior?

Both tabs are "upgraded" (` mdc-ripple-upgraded`, etc.).

### What is the actual behavior?

The first tab is "upgraded", but the second is not.

### Any other information you believe would be useful?

2017-12-22 14_55_39-localhost_8080_ sh7hfjvi1kzt5y5xqjsntlidnz03_u0wfkx6bgvpd3stqevxi69ex9pm1

vdkdev:
Fixes #67

### DIFF
--- a/src/Tabs/index.js
+++ b/src/Tabs/index.js
@@ -49,6 +49,25 @@ export const TabBar: React.ComponentType<TabBarPropsT> = withMDC({
     if (!props || nextProps.activeTabIndex !== props.activeTabIndex) {
       api.activeTabIndex = nextProps.activeTabIndex;
     }
+  },
+  didUpdate: (props, nextProps, api, inst) => {
+    if (!api) return;
+
+    const childrenDidChange =
+      props &&
+      props.children &&
+      nextProps &&
+      nextProps.children &&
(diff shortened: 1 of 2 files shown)


vdkdev:
I've fixed the check to use the key property array literal for diffing. It's probably an overkill to have it there. Also, the event handlers were not being released so MDCTabFoundation->destory is now called before the TabBar component is reinitialized.

and updated the story to test the dynamic nature of it.

jamesmfriedman:
Thank you so much for this! I’ll dig in later today and make sure this gets in the next release.

PR #76, opened 2018-01-04
Move react and react-dom to peerDependencies

alexeyraspopov:
It is better to have those two defined as [`peerDependencies`](https://docs.npmjs.com/files/package.json#peerdependencies). Spotted the issue when tried to use [bundlephobia.com](https://bundlephobia.com/result?p=rmwc).

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,10 @@
     "material-components-web": "^0.27.0",
     "prop-types": "^15.6.0"
   },
+  "peerDependencies": {
+    "react": "^16.0.0",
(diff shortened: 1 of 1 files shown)


jamesmfriedman:
Great catch, I never actually knew about peerDependencies in package.json. 

One request before I merge, RMWC supports React 15 and up, so can you make it 15.x.x and up? I’m not sure of the exact notation, I think it might just be ^15.x.x.

alexeyraspopov:
@jamesmfriedman, updated with condition

Issue #80, opened 2018-01-05
Want to be a Collaborator? Comment here.


jamesmfriedman:
Hi all, I started this project out of a an unmet need and opportunity I saw to fill it. RMWC has been getting some good traction and feedback, and I'd like to find some people to help take it to the next level.

If you're interested, comment on this issue and lets get a conversation going.

izakfilmalter:
I would like to help. In the middle of trying to get my startup off the ground. Once it is launch I can def help for about 5 hours a week.

j-o-d-o:
I'd have some time I could spend.

jamesmfriedman:
Thanks @izakfilmalter, let me know when your time gets freed up.

@j-o-d-o, I just added you as a collaborator.

jrajav:
I am interested but can't commit regular time. I will try to take a look at a few issues at least.

jamesmfriedman:
Thanks @jrajav, every little bit helps

GideonFlynn:
I would love to contribute to this awesome project. I have a flexible schedule and can start with about 3 hours a week.

jamesmfriedman:
Thanks @GideonFlynn, how about checking out the massive #88 PR?

bernharduw:
I could get a project done within 3 days thanks to rmwc! I want to give back (although I won't be able to contribute too much for now). I'd like to do those theming issues, #89 and maybe @90, because this was what my project was about.

jamesmfriedman:
Thanks @bernharduw, happy to accept PR's for either if you get the time to work on it. Just chime in on those issues.

ALMaclaine:
I'm very interested in helping. I'm still somewhat new, just about 3 years.  I'm very interested in Material Design, React and helping maintain a professional framework. Might need some help getting jump started on understanding the code base but I believe I could be useful in some capacity.

jamesmfriedman:
Thanks @ALMaclaine. Feel free to look through the issues and submit any PRs. Theres a growing group here for guidance and help.

ALMaclaine:
@jamesmfriedman will do, thank you.

ricardobrandao:
Hi @jamesmfriedman. It's good to see that RMWC has been evolving over time! 👍 

I was doing my own version of the react material components but I'm now willing to give a try to RMWC. Therefore, I'll gladly help with the project. Right now I have a lot on my plate but I'll try to dedicate some time to this.

jamesmfriedman:
[...]

PR #81, opened 2018-01-06
fix(drawer): list items don't show name and icon in story

j-o-d-o:
Change to the correct key `displayName` and add the icon as `<ListItemStartDetail>`


jamesmfriedman:
curious @j-o-d-o, are you able to merge after an approved review? Not sure how this works on public repos.

j-o-d-o:
Nope, it says only users with write access can merge pull requests.

jamesmfriedman:
Speaking of which, any interest in #80?

jamesmfriedman:
Ah, just saw your comment!

PR #85, opened 2018-01-08
Move react-scripts dep to devDependencies

jrajav:
Fixes #85

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,9 +29,7 @@
   },
   "dependencies": {
     "classnames": "^2.2.5",
-    "material-components-web": "^0.27.0",
-    "react": "^16.2.0",
-    "react-dom": "^16.2.0"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -31,8 +31,7 @@
     "classnames": "^2.2.5",
     "material-components-web": "^0.27.0",
     "react": "^16.2.0",
```
jamesmfriedman: @jrajav can you do me a favor and also remove react and react-dom from dependencies? They made it in there by accident. These are covered in peer dependencies, so they don't have to be moved, you can just safely delete them.

jamesmfriedman:
#84

package.json
jrajav: @jamesmfriedman Done!

jamesmfriedman:
Thanks @jrajav

PR #98, opened 2018-01-12
Remove reliance on Reflect

### LINKED ISSUE #95: Document Reflect polyfill needed for IE11 and below
`Reflect` isn't supported in IE11 and below.

JetFault:
Closes #95

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -24,6 +24,7 @@
       }
     ],
     "react/prop-types": "off",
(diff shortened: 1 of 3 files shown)


jamesmfriedman:
Looks good @JetFault , thanks for the PR! This will fail linting though. Can you disable the ESLint rules for those lines? I'm not sure what your setup is, but eslint should be installed as a dev dependency.

PR #102, opened 2018-01-15
Add sandbox to issue template

j-o-d-o:
I created a Sandbox for reproducing issues and bugs. Code always say more than a thousand words ; )


jamesmfriedman:
Awesome, does that mean the new UMD bundle is working now?

j-o-d-o:
I wanted to use Codepen (like mdc does), but it still doesn't work. Now there are new errors when rmwc creates mdc components. I opted against going down that rabit hole and chose another sandbox tool which supports es6 import and npm modules (https://codesandbox.io in this case). 
I think it is much more comfortable and maps a real react application much better and it is easier to include other npm dependencies.

jamesmfriedman:
Thanks for setting this up!

Issue #104, opened 2018-01-17
Select float above not set when value is null


jamesmfriedman:
When Select mounts with a value, it should automatically set the label to "float-above"

j-o-d-o:
I can't reproduce this bug. When there is a Select with a hardcoded value property, that would mean it "mounts with a value" right?

j-o-d-o:
https://github.com/jamesmfriedman/rmwc/blob/633d9c846c428009103f99b2060fcd9d632e768e/src/Select/index.js#L27

Don't know if you already looked into it. But guess this line is the problem. null equal false in if clause. 
Question is, should it be the default behavior to show the first selection in the list if the value is null?

Here is a usecase: I have something i want the user to select. I write `this.state = { my_selection: null }` and `<Select value={this.state.my_selection} ... />`. Now the user thinks: "Ah cool, I wanted to have the first selection anyway and presses submit". But `this.state.my_selection` is still `null`. 

Thoughts?

jamesmfriedman:
Yeah thats where I landed as well. Didn't get a chance to dig in any farther. It has something to do with the default selection. Null is definitely not a valid option for a value, but it's defaulting to the first option in options.

j-o-d-o:
Okay, so the issue is actually to set the placeholder or the label (non floating) if value is null?

jamesmfriedman:
the actual issue seems to be that `undefined = no selected value` from the select , and `null = first value selected`

jamesmfriedman:
Thanks for the PR (And for scrubbing the issues today). I really don't know what the "right" or "expected" solution is. I'll pull the PR tomorrow and play around with it.

PR #105, opened 2018-01-18
fix(Select): Don't show first option if value = null

j-o-d-o:
PR to the discussion in #104. Need to figure out the desired behavior first. I will change this PR in case that changes to something other than showing the label / placeholder for value = null.

### DIFF
--- a/src/Select/index.js
+++ b/src/Select/index.js
@@ -175,7 +175,10 @@ export const Select = withMDC({
       isFirstRun ||
       placeholderDidChange
     ) {
-      const newIndex = api.options.indexOf(api.nameditem(nextProps.value));
+      // if value = null, MDC will show the first option. This behaviour is not desired
+      let value = nextProps.value === null ? undefined : nextProps.value;
+
+      const newIndex = api.options.indexOf(api.nameditem(value));
       api.selectedIndex =
         newIndex === -1 && nextProps.placeholder ? 0 : newIndex;
 
--- a/src/Select/index.js
+++ b/src/Select/index.js
@@ -176,7 +176,7 @@ export const Select = withMDC({
       placeholderDidChange
     ) {
       // if value = null, MDC will show the first option. This behaviour is not desired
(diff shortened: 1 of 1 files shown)


src/Select/index.js
```diff
@@ -175,7 +175,10 @@ export const Select = withMDC({
       isFirstRun ||
       placeholderDidChange
     ) {
-      const newIndex = api.options.indexOf(api.nameditem(nextProps.value));
+      // if value = null, MDC will show the first option. This behaviour is not desired
+      let value = nextProps.value === null ? undefined : nextProps.value;
```
jamesmfriedman: Use `const` when the value doesn't change.

jamesmfriedman:
Just a heads up @j-o-d-o since you are a collaborator, you should be able to make branches directly inside of RMWC. When you do, just use the format yourname/your-fix.

Played around with this some, it definitely seems to be the desired behavior. I tested it against other falsey values and it worked just fine. I found out why in MDC this actually happens.

```javascript
function nameditem(key) {
  // NOTE: IE11 precludes us from using Array.prototype.find
  for (var i = 0, options = this.options, option; option = options[i]; i++) {
    if (option.id === key || option.getAttribute('name') === key) {
      return option;
    }
  }
  return null;
}
```

When option.id === key evaluates to false (the value wasn't found) it tries to fall back to option.getAttribute('name') which defaults to `null` when there is no "name" attribute on the DOM item. Definitely an odd quirk, but at least it makes some sense now.

Personally, I'd like to refactor this at some point to be completely value driven instead of using the namedItem api which looks at the DOM, but for now other than the comment to change `let` to `const`, It's fine to merge in and definitely fixes the issue.

j-o-d-o:
Yeah, looks like that's it, good catch. Not sure if this is on purpose or not in mdc. I will check this out in the mdc chat and possibly create and issue in the mdc repo.

jamesmfriedman:
I think it’s fine I’m the MDC code. Since mdc deals primarily with the DOM I would expect this kind of behavior.

PR #107, opened 2018-01-19
fix(tabBarScroller) : Trigger context update after TabBarScroller creates TabBar api.

vdkdev:


jamesmfriedman:
Just a heads up @vdkdev, you are a collaborator on the project. You can now create, push, and pull directly from RMWC, the only protected branch is master.

PR #115, opened 2018-01-23
build(RMWC): Add missing "react" and "react-dom" (dev)dependencies for local development

ricardobrandao:
Following the development guidelines won't work because react and react-dom are listed as peerDependencies and therefore are not downloaded. 

Adding these dependencies as devDependencies will not mess with the production build and will ease up the contributors' life.


jamesmfriedman:
I actually just updated the Contributing readme to detail this exact problem here: e927698b30f22b8c0d22839778b72f53231e229f

The "correct" thing to do is have it listed as a peer dependency and install it manually. If it is set as a devDependency I *think* it will install its own local version and ignore any other version that was installed.

With that said, I think you are right, it really shouldn't cause any conflicts with anything. Any more thoughts before we merge this?

jamesmfriedman:
Just checked out a bunch of other React libraries and this is how they handle it. Merging now.

PR #117, opened 2018-01-23
v1.1.0 - MDC 29

jamesmfriedman:
Not many breaking changes from MDC this time (thankfully).

- Removed the multi-select
- documented ToolbarFixedAdjust
- documented an additional case for Enzyme testing setup
- Select now re-inits appropriately when switching between cssOnly and fully featured
- Added a global dark theme to the docs that you can toggle on and off as well as some additional docs cleanup

Anything else we should add to this release?


jamesmfriedman:
@j-o-d-o @izakfilmalter I just added a more fluid fix for the testing issue people have been experiencing. Check out the latest commit. Would be great if someone could verify it works for them as well.

jamesmfriedman:
Can anyone else get some eyes on this today? The changes are really minimal as far as breaking changes go, but there are a good amount of small fixes that have made their way in from the community.

I also want to make sure the testing polyfill will work for everyone. It at least makes life easier :).

@alecchisi @JetFault @UFOMelkor @alexeyraspopov @jrajav @j-o-d-o @izakfilmalter

izakfilmalter:
I can look at it tomorrow. Today I'm smashed. Still need to figure out
tabbar as well.

jamesmfriedman:
@izakfilmalter  Yeah, sorry you're still having those issues. Hit me with a repro of it if you get a chance and I'll take a look at it tonight.

j-o-d-o:
LGTM

JetFault:
I tested it on our site (we heavily use text fields) and all the text field changes look great. I also verified 9d886b6 was working correctly. 

Tests all completed successfully.

This also fixed a bunch of mdc textfield issues I had with mdc which is nice :)

CONTRIBUTING.md
```diff
@@ -13,7 +13,7 @@ You'll need a recent version of [nodejs](https://nodejs.org/en/) to work on RMWC
 > **NOTE**: If you expect to commit updated or new dependencies, please ensure you are using npm 5, which will
 > also update `package-lock.json` correctly when you install or upgrade packages.
 
-Once node is installed, simply clone our repo (or your fork of it) and run `npm install`
+Once node is installed, simply clone our repo (or your fork of it) and run `npm install`. You'll notice you have to install react and react-dom with the no-save option. These are peer dependencies that are not installed automatically.
```
ricardobrandao: This does not apply anymore, right?

izakfilmalter:
Testing in my app this morning. Will also try to add a fix for the `TabBar` stuff.

CONTRIBUTING.md
j-o-d-o: React is still as a peer dependency in the package.json as I see it. So installing react is still needed right?

jamesmfriedman: Someone put in a PR that fixed it. Looked at other libraries, they still include React as a dev dependency which shouldn’t cause any problems when being used in a normal context with npm install.

jamesmfriedman:
@izakfilmalter it would be awesome to simplify that clause. Basically the tab bar needs to re-init if it’s children change, but we have all the problems of checking object identity with that. 

I’m bummed you can’t repro it :/

CONTRIBUTING.md
j-o-d-o: Ah, was looking at the v1.0.2 tag ... sorry.

izakfilmalter:
Gonna pull the patch and then run my tests.

PR #138, opened 2018-02-04
feat(GridList): add GridTileIcon

mattmichler:
Adds a `<GridTileIcon>` component to take advantage of the `withIconAlignStart` prop on `<GridList>`. It uses the existing `<Icon>` component and adds the MDC class `mdc-grid-tile__icon` to the element.


jamesmfriedman:
Looks goo, thanks for this Matt! Will put it out as a patch today.

Issue #139, opened 2018-02-06
MDC 30


jamesmfriedman:
- [x] Rename SimpleMenu to Menu and all relevant classes
- [x] Create an RMWC simple menu that can automatically handle the open state
- [x] It seems like dark themes have been removed altogether? Validate this and adjust accordingly.
- [x] Add a chip component
- [x] remove cssOnly select
- [x] add box version of Select and validate non-box version works appropriately
- [x] mdc-text-field__bottom-line -> mdc-line-ripple
- [x] mdc-checkbox__checkmark__path -> mdc-checkbox__checkmark-path
- [x] Cards layouts have been completely gutted, respond accordingly
- [x] addition of mdc-button__icon
- [x] return of the mini fab... mdc-fab--mini 
- [ ] match minor version to MDC for next release.

j-o-d-o:
What's the reason for removing all of the "css-only" components? Seems like something lot's of users actually need. Especially for a select component e.g. on mobile...

j-o-d-o:
Regarding Textfield `mdc-text-field__bottom-line`. I think it is `mdc-line-ripple` now and it will be a new component in rmwc?

j-o-d-o:
Also, how can I add breaking changes into the changelog? Regarding the change of SimpleMenu -> Menu?

jamesmfriedman:
Thanks for jumping on this one @j-o-d-o!

- I **think** the reason they are removing the css-only components is because they can't guarantee styling or functionality across browsers. In the instance of select, it makes a lot of sense, but the cssOny one never matched up exactly with the JS one. Whats your thoughts on supporting a cssOnly select directly in RMWC? It would be the first attempt at preventing a breaking change but might also involve doing some stying which RMWC doesn't currently include.

- You are correct on mdc-line-ripple. Apparently they called it bottom line first and then changed the name

- for the changelog, follow the patterns [here](https://www.npmjs.com/package/generate-changelog). For breaking you can simply type `breaking(Select): my description...`. I know it's annoying, but if you can amend that specific commit title then the changelog will be generated appropriately.

- Lets keep that mdc30 branch open and keep working off of it until we get all of the features in there.

- Maybe we can convince the people over at material-components-web to have a blog related to why they are doing the things they are? It would really help with dramatic changes like this. I don't know how to get them to do that though other than opening an issue that looks like I'm just complaining. Either way, I asked them to keep the docs updated politely and that hasn't happened at all :(.

[...]

PR #140, opened 2018-02-06
MDC 0.30.0

j-o-d-o:
#139


src/Select/index.js
```diff
@@ -67,10 +80,9 @@ type SelectPropsT = {
   /** Placeholder text for the form control. */
   placeholder?: string,
   /** Disables the form control. */
-  disabled?: boolean,
-  /** Makes a cssOnly select. */
-  cssOnly?: boolean
```
izakfilmalter: RIP

jamesmfriedman: Hahaha. I could tell by your cod examples you loves them. Shouldn’t be hard to make a wrapper for.

izakfilmalter: It's more that menu has issues when inside something that is transformed or has overflow hidden. Which is like 3/4 of my mobile interface.

jamesmfriedman:
Alright, looks like we have a release. Thanks for you help @j-o-d-o and @izakfilmalter. Merging to master and doing one last check before the deploy.

PR #144, opened 2018-02-11
Sync validality status with TextField Foundation

lo-tp:
Beforehand, the invalid prop in the react component is not synced with TextField Foundation


jamesmfriedman:
Thanks so much for this!

Issues like this are giving me more reason to look into a direct foundation implementation but this is great. Will put it out as a patch before the next release.

PR #145, opened 2018-02-12
Bugfix: issues 143

lo-tp:


jamesmfriedman:
Thanks for this very basic, albeit very logical fix. I'm trying to think of any reason where we wouldn't want a PureComponent and I can't.

@j-o-d-o @izakfilmalter any thoughts on this?

Issue #147, opened 2018-02-20
TextField does not change aspect when set as "disabled"


alecchisi:
### What RMWC Version are you using?

>RMWC 1.2.0

### What browser(s) and React Version is this bug affecting?

Any browser

### What build system are you using?

> Webpack + Create-React-App

### What are the steps to reproduce the bug?

Adding a `<TextField disabled>` element to a React JSX component.

### What is the expected behavior?

The Textfield should be rendered as disabled, by adding the `mdc-text-field--disabled` to the wrapping `<div>`

### What is the actual behavior?
Inspecting the DOM, you cannot see the `mdc-text-field--disabled`  added.

j-o-d-o:
Good catch! PR should fix this.

j-o-d-o:
Resolved with the release of 1.2.1

PR #148, opened 2018-02-20
fix(textfield): add mdc class on disabled textfields

j-o-d-o:
When using the disabled property on textfield `mdc-text-field--disabled` should be added as a className. Fixes issue #147


j-o-d-o:
Yes, sorry. 
I first tried it and adding the class was enough to prevent the user from using the Textfield. But the MDC documentation says you should do both, add the className and add disabled to the input element. 
https://material.io/components/web/catalog/input-controls/text-field/

jamesmfriedman:
Awesome, thanks for the quick turn around @j-o-d-o .

I'll get this merged and put out a patch

Issue #150, opened 2018-02-21
MDC 31


jamesmfriedman:
This week on MDC, nice and easy!

- Card: Add Primary action element
- Chips: Leading and Trailing icons

jamesmfriedman:
Fixed with #151

PR #151, opened 2018-02-23
MDC 31

jamesmfriedman:
This one was uncharacteristically easy. #150

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,12 +5,19 @@
 
 # RMWC - React Material Web Components
 
-A React (15 / 16) wrapper for Material Design (Web) Components v0.30.0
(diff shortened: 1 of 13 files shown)


jamesmfriedman:
That seems like a good thing to add programmatically. Maybe make a wrapper component with that logic and then pass it to a ChipIconRoot?

j-o-d-o:
Sounds like a plan. Yet, the user should still be able to not use tabindex and role.

j-o-d-o:
But I guess we can also merge this right now and publish and keep that extension in mind for the future.

jamesmfriedman:
Thanks @j-o-d-o. I'll get this merged and deployed. Nice to have an easy week :).

PR #165, opened 2018-03-15
MDC 32

jamesmfriedman:
The biggest change here is the beginning of foundation adapter implementations. Please test this as thoroughly as you possibly can.


jamesmfriedman:
https://github.com/material-components/material-components-web-react/issues/4

PR #183, opened 2018-04-05
MDC 34

jamesmfriedman:
Excited about this release, things are getting cleaner and cleaner.

@j-o-d-o you're the most well versed in this stuff. Can you give this your seal of approval?

also @jrajav, I know you had a handful of issues recently, wanted you to take a look as well if you can.


ginavasiloff:
This looks good! Was able to compile and test locally.

package.json
```diff
@@ -72,8 +77,22 @@
     "react-app-rewired": "^1.4.0",
     "react-dom": "^16.2.0",
     "react-hot-loader": "^3.1.3",
-    "react-markdown-loader": "git+https://github.com/jamesmfriedman/react-markdown-loader.git#master",
+    "react-markdown-loader":
+      "git+https://github.com/jamesmfriedman/react-markdown-loader.git#master",
```
j-o-d-o: A particular reason for the new lines for some key values? Linting? Auto format? To me it looks nicer if everything is a one liner, especially since it doesn't really save much horizontal space anyway in most cases.

src/TextField/index.js
```diff
@@ -148,133 +154,118 @@ type TextFieldPropsT = {
...
-      api.foundation_.setValid(!nextProps.invalid);
-    }
-  },
-  onMount: (props, api, inst) => {
-    api.foundation_.setValid(!props.invalid);
+export class TextField extends withFoundation({
+  constructor: MDCTextField,
+  adpater: {
+    addClass: addClass(),
+    removeClass: removeClass(),
```
j-o-d-o: Can you think of any component that does not have `addClass()` and `removeClass()`? Maybe putting this into the base class would be a good idea?

src/TextField/index.js
```diff
@@ -148,133 +154,118 @@ type TextFieldPropsT = {
...
+    getValue: () => {}
```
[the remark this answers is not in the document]
jamesmfriedman: Good catch. I also misspelled adapter which means this wasn't getting picked up at all. Fixed.

src/TextField/index.js
jamesmfriedman: Not every adapter has to use it. I'm only keeping it here so there isn't any magical overriding of the adapter methods, what you see in these files is what you get. Otherwise your source of truth is whats in this file, whatever MDCFoundation would do, and whatever the MDC class is doing. Just seems harder to reason about.

jamesmfriedman:
@j-o-d-o yeah it's a bummer about the select. But on the other hand, performance on mobile was so terrible before that I think this is the lesser of two evils.

PR #214, opened 2018-04-17
fix(Foundation): Component destroy was not being copied onto the new …

jamesmfriedman:
…prototype and called resulting in memory leaks.

@j-o-d-o This is a critical issue that I ran into late last night. Never showed up in Chrome, but mobile safari was crashing left and right.

The short of it:
- I wasn't copying the destroy method onto the new prototype
- The components destroy method was never being called
- And foundation component that has any logic in its constructor has to have that logic duplicated in RMWC in the initialize function. This seems few and far between but inconsistent on their end. We can't reference DOM elements until the component has mounted anyways, so even if I could call their constructor, this.root_ would be undefined.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -651,6 +651,126 @@
       "integrity": "sha512-+UWvVSr3KrnDUa1JXv8cdirzcM9jrjcthgWajmvU6TlBnzvSqF6oHLeDTGOcNPp4zGg12kSCAjB3dnNdVEkZOw==",
       "dev": true
     },
+    "@storybook/codemod": {
+      "version": "3.4.1",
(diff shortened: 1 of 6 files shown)


jamesmfriedman:
That’s exactly what it’s for. 

The ripple specifically doesn’t clean itself up immediately when calling destroy, it has a few frames where it is still doing things like removing styles and computing things. 

Without waiting you run into instance.root_ is undefined errors which will crash iOS. I didn’t want to make this number any higher than it needed to be since it’s keeping the dom nodes around in memory and your components are likely being cleaned up because you switched routes or views in your app which means you need the memory.

Issue #210, opened 2018-04-16
MDC 35


jamesmfriedman:
Goals of the release are to focus on the foundation implementation, testing, and reliability.

MDC Breaking Changes:
- [x] Button: rename stroked to outlined
- [x] Card: rename stroked to outlined
- [x] Typography: renamed most types
- [x] Audit theme changes

New MDC Components:
- [x] Shape
- [x] Image List
- [x] Top App Bar

Foundation Implementations
- [x] Switch #190 
- [x] Checkbox #191 
- [x] Radio #192 
- [x] Form Fields #194 
- [x] Menu #196 

Other Goals
- [x] SSR tests for the new foundation components #171
- [x] 100% flow code compliant, more work on #159

jamesmfriedman:
Released 1.6

PR #221, opened 2018-05-02
MDC 35

jamesmfriedman:
Upgrading to the latest release of MDC. Please see the associated issue #210 for a work list of changes.


src/Checkbox/index.js
```diff
@@ -72,59 +82,100 @@ export type CheckboxPropsT = {
...
+  }
+
+  render() {
+    const {
+      label = '',
+      id,
+      children,
+      checked,
+      indeterminate,
+      apiRef,
```
j-o-d-o: Does it make sense to still include `apiRef`? e.g. in the select component, we got rid of it.  (Same question for radio button)

jamesmfriedman: It actually still exists for now since I'm not sure if anyone was using it. See this line https://github.com/jamesmfriedman/rmwc/blob/mdc35/src/Base/MDCFoundation.js#L181

I didn't know this about React, but it turns out using a "ref" on a class component and not a Dom element actually gives you a reference to the class. So for all Foundation based components since they are merged with the original vanilla JS class, doing something like this would identical:

```
<Checkbox apiRef={api => {}} />
<Checkbox ref={api => {}} />
```
The only difference is timing. The standard React ref will fire during render where as the apiRef will only fire once the component mounts. As soon as all of the components are moved over to the new pattern, we can add a deprecation notice in the code and remove.

j-o-d-o: Ah okay, actually that's what I did for all my selects that I am using: switching from `apiRef` to `ref`, cause on select the console says it does not recognize `apiRef`. 
But guess a big breaking change in the end for all components is the way to go

jamesmfriedman: Which means on select I forgot to remove it :)

jamesmfriedman: Thanks for the review @jodo and @amaesato

PR #238, opened 2018-05-30
Fix IE11 support for MDCFoundation

JetFault:
Symbols don't exist on IE11

### DIFF
--- a/src/Base/MDCFoundation.js
+++ b/src/Base/MDCFoundation.js
@@ -22,7 +22,7 @@ const requestFrames = (
 /** Copies all known properties from source to target. This is being used in here for class merging. */
 const copyProperties = (target, source) => {
   const allPropertyNames = Object.getOwnPropertyNames(source).concat(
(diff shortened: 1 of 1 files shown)


jamesmfriedman:
Sorry I didn't see this sooner @JetFault . It looks like you default to an empty array. This function is whats actually copying the class properties / methods over. I'm curious how this would work at all in this example.

JetFault:
My thinking is if `Object.getOwnPropertySymbols` doesn't exist, then there isn't any Symbols support, so it should return `[]`.

JetFault:
Also, I didn't find any usage of Symbols in mwc-web or rmwc, and by default there are no Symbols on an object. 
image

jamesmfriedman:
You're totally right, and I definitely misread the code. I'm not sure that needs to be in there at all, that was a copy pasta from stack overflow on runtime merging of classes! I'll get a patch out.

PR #261, opened 2018-07-05
fixes for missing dev dependencies and incorrectly defined flow types

mutsys:
The first commit just contains the updates to `package.json` that are needed to perform a build.

The second commit contains a series of fixes to some of the `flow` type definitions. While they were acceptable for JS projects their translation to TS during the build resulted in erroneous TS type definitions. Some elements had props with missing or improperly defined attributes . `tsc` is quite a bit more strict about type checking than `flow` apparently.

### DIFF
--- a/src/Drawer/index.js
+++ b/src/Drawer/index.js
@@ -57,11 +57,11 @@ export const DrawerContent = simpleTag({
  ***************************************************************************************/
 export type DrawerPropsT = {
   /** Opens or closes the Drawer. */
-  open: boolean,
+  open?: boolean,
   /** Callback that fires when the Drawer is closed. */
-  onClose?: (evt: CustomEventT) => mixed,
+  onClose?: (evt: CustomEventT<void>) => mixed,
   /** Callback that fires when the Drawer is opened. */
-  onOpen?: (evt: CustomEventT) => mixed,
+  onOpen?: (evt: CustomEventT<void>) => mixed,
   /** Makes a permanent drawer. */
   permanent?: boolean,
   /** Makes a persistent drawer. */
@@ -115,7 +115,7 @@ const slidableDrawerFactory = (MDCConstructor, displayName) =>
           syncFoundationProp(
             nextProps.open,
             this.open,
-            () => (this.open = nextProps.open)
+            () => (this.open = nextProps.open || false)
           );
       });
     }
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "eslint-plugin-import": "^2.11.0",
     "eslint-plugin-jsx-a11y": "^5.1.1",
     "eslint-plugin-react": "^7.7.0",
+    "extract-text-webpack-plugin": "^3.0.2",
     "flow-bin": "^0.74.0",
     "generate-changelog": "^1.7.0",
     "normalize.css": "^7.0.0",
@@ -92,7 +93,9 @@
     "react-router-dom": "^4.2.2",
     "react-scripts": "^1.1.4",
     "storybook": "^1.0.0",
-    "typescript": "^2.9.2"
+    "sw-precache-webpack-plugin": "^0.11.5",
+    "typescript": "^2.9.2",
+    "webpack-manifest-plugin": "^2.0.3"
   },
   "jest": {
     "resolver": "./scripts/jest-resolver.js",
--- a/src/Base/index.js
+++ b/src/Base/index.js
@@ -14,19 +14,4 @@ export { randomId } from './utils/randomId';
 
 export { withFoundation, syncFoundationProp } from './withFoundation';
 
-export type CustomEventT = {
-  bubbles: boolean,
-  cancelBubble: boolean,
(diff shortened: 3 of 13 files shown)


jamesmfriedman:
Hi @mutsys, thanks for the PR. A few questions for you:

- What failures were you having with the deps you added to package.json? The dependencies you listed out get automatically installed by create-react-app (react-scripts specifically). There shouldn't be a need to track them manually since they are not direct dependencies. They are also only required to build the docs

- Can you be more specific about the issues you were hitting with Typescript? I just want to understand better.

- I'll dig into this PR a bit more tomorrow, looks good so far though.

src/Drawer/index.js
```diff
@@ -115,7 +115,7 @@ const slidableDrawerFactory = (MDCConstructor, displayName) =>
           syncFoundationProp(
             nextProps.open,
             this.open,
-            () => (this.open = nextProps.open)
+            () => (this.open = nextProps.open || false)
```
jamesmfriedman: should probably be `this.open = !!nextProps.open`

mutsys:
Regarding the apparently missing dev dependencies, here are the steps to reproduce:

```bash
git clone https://github.com/jamesmfriedman/rmwc.git
cd rmwc
yarn install
yarn build
.
.
.
Error: Cannot find module 'extract-text-webpack-plugin'

yarn add extract-text-webpack-plugin --dev
yarn build
.
.
.
Error: Cannot find module 'webpack-manifest-plugin'

yarn add webpack-manifest-plugin --dev
yarn build
.
.
.
Error: Cannot find module 'sw-precache-webpack-plugin'

yarn add sw-precache-webpack-plugin --dev
yarn build
.
.
.
Compiled successfully.
```

mutsys:
Regarding the issues with Typescript and the definitions of `rmwc` component prop interfaces...

The core of the issue is due to the way React deals with what it calls "intrinsic attributes". Even though your props interface does not explicitly extend a "base" interface providing these attributes (such as `children`, `key`, etc...), React will enhance your component at runtime and adds additional interfaces to your props.

Take a look at the main React type definition file (`@types/react/index.d.ts`):
* `IntrinsicAttributes` - adds `key`
* `IntrinsicClassAttributes` - adds `ref`
* `HTMLAttributes` - adds common HTML attributes such as `id`, `style` and `className`

Any attributes found in these three interfaces should never be explicitly added to a prop interface in Typescript, however many of the `rmwc` do define some of these (sporadically, inconsistently and in some cases incorrectly) on a number of components. As I mentioned previously, you should not have to extend any of those three "base" prop interfaces in order to define their attributes on a component, however, if the `SimpleTagPropsT` does not explicitly extend `HTMLAttributes` building `rmwc` fails with a variety of errors such as:

```
TextField/index.tsx:217:16 - error TS2339: Property 'id' does not exist on type '{ value?: string | number | undefined; required?: boolean | undefined; type?: string | undefined;...'.

217       id: rest.id || randomId('text-field')
                   ~~
```

Very puzzling, this should not be happening and it should not be necessary for `SimpleTagPropsT` to extend `HTMLAttributes` in order to prevent these errors. I think this is a symptom of a larger problem in general, and that is I think the `rmwc` `flow` type definitions contain a number of errors that don't seem to impact `rmwc` usage from Javascript since it doesn't go through a type checker and transpiler that depend on the definitions. But when those erroneous definitions get converted to Typescript they result in erroneous type definitions and `tsc` fails.

There are a number of other issues similar to this one (such as the definition of `CustomEvent`) that I addressed as well. All of this leads me to ask the following question:

If type definitions are important to you (as they should be), then why not switch the project from JS/Flow to Typescript?

* It can be done incrementally, a component at a time.
* Tools exist to extract `flow` type definitions from Typescript.
* Editor/IDE support for Typescript "as-you-type" type checking is excellent and much faster than the equivalent for `flow`.
* Typescript has gained much wider adoption than `flow`.

The `rmwc-typed` project (to which I have contributed) currently contains a set of well vetted type definitions for `rmwc` that could used as a means to jump-start that effort. Merge those definitions into `rmwc` proper, ditch the flow -> *.d.ts step in the build and then each component could be converted one at a time with zero impact on dependent projects. Just a thought...

jamesmfriedman:
@mutsys I ended up cherrypicking your changes and correcting them as needed in commit https://github.com/jamesmfriedman/rmwc/pull/268/commits/7b4840d94e04b3cd95eea6e19b855143261762e0
They're in an open PR for #268 

A few things
- The missing dev dependencies is a Yarn related issue and does not happen with NPM. According to some things I read, some versions of yarn don't automatically install your dev dependencies. On such thread https://github.com/yarnpkg/yarn/issues/2739.
- Your changes might have fixed Typescript, but they broke flow :). I've made a few tweaks to get them to work for both.
- I'm going to go ahead and close this, but I'd love to get your validation on #268.

mutsys:
Hi James,

   Just checked it out and ran a few quick tests. Everything looks good to
me.

Mike

PR #271, opened 2018-07-16
Fix ThemeProvider example in docs

schliflo:
Currently the ThemeProvider example code shown in the TopAppBar of the docs is broken. It's missing the secondary braces and still imports from `rmwc/ThemeProvider` which - i guess - was moved to `rmwc/Theme`.


jamesmfriedman:
Thanks, great catch! Sorry you had to dig through my garbage docs code :)

PR #273, opened 2018-07-22
fix(docs): renamed docs menu and url name Styling to Styling and Theming

CoVoCre:
renamed docs menu and url name Styling to Styling and Theming, renamed styling.md file and the imports in menuContent.js


CoVoCre:
Oh wait, maybe it isn't right, I should've done docs(Styling):.... I'll try it again !

PR #274, opened 2018-07-22
Small changes to docs Styling page

CoVoCre:


jamesmfriedman:
Thanks for the PR @CoVoCre.

RMWC uses prettier to ensure consistent styling. You'll have to make sure it is enabled in your editor. I can't accept this PR as is because the majority of it is your IDE overwriting indentation, and single to double quotes.

CoVoCre:
@jamesmfriedman thing is, I have prettier installed. I now set double to single quotes, installed eslint but still get some format changes...
I tried some new things, it hasn't fixed every format changes but a lot...

PR #290, opened 2018-08-24
Bugfix/typedefs fixes

shawnatxapp:
Hey James.

I am working on incorporating the IconButton into our project and ran into some TS errors:

```
node_modules/rmwc/Icon/index.d.ts:59:5 - error TS2416: Property 'render' in type 'Icon' is not assignable to the same property in base type 'PureComponent<IconPropsT, {}>'.
  Type '() => ReactNode' is not assignable to type '() => false | Element | null'.
    Type 'ReactNode' is not assignable to type 'false | Element | null'.
      Type 'undefined' is not assignable to type 'false | Element | null'.

59     render(): React.ReactNode;
       ~~~~~~


node_modules/rmwc/Provider/index.d.ts:51:5 - error TS2416: Property 'render' in type 'RMWCProvider' is not assignable to the same property in base type 'Component<RMWCProviderOptionsT, {}>'.
  Type '() => ReactNode' is not assignable to type '() => false | Element | null'.
    Type 'ReactNode' is not assignable to type 'false | Element | null'.
      Type 'undefined' is not assignable to type 'false | Element | null'.

51     render(): React.ReactNode;
       ~~~~~~
```

It seems to me that the first is due to the fact that the Icon component's render method could [potentially return null](https://github.com/jamesmfriedman/rmwc/blob/master/src/Icon/index.js#L107)

The other seems to be because the Provider component's children prop could [potentially be  undefined](https://github.com/jamesmfriedman/rmwc/blob/master/src/Provider/index.js#L23). Which is technically correct for the React.Component type but not applicable to the Provider.

This addresses that and from what I can tell solves the issue.

Let me know if I am off base or if there is a better way to fix this.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,10 +20,10 @@
       "dev": true,
       "requires": {
         "@babel/types": "7.0.0-beta.44",
-        "jsesc": "2.5.1",
-        "lodash": "4.17.10",
-        "source-map": "0.5.7",
-        "trim-right": "1.0.1"
+        "jsesc": "^2.5.1",
+        "lodash": "^4.2.0",
+        "source-map": "^0.5.0",
+        "trim-right": "^1.0.1"
       },
       "dependencies": {
         "jsesc": {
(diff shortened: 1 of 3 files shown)


shawnatxapp:
Aces. Love the IconButton! The switch functionality is :fire:

shawnatxapp:
Aces. Love the IconButton by the way. The switch functionality is 🔥

PR #303, opened 2018-09-08
fix(TopAppBar): Corrects alignStart style for TopAppBarSection

cyberhuman:
This PR corrects a typo in TopAppBarSection's className when alignStart=true so that proper style is applied.


jamesmfriedman:
Thanks!

PR #309, opened 2018-09-13
fix(TextField): Textarea is including margin-top.

arpontes:
When textbox are marked as boxed or outlined, this margin is not applied, but it should not be applied also when textarea is set.
We can see the incorrect margin in demo of textareas, when the control receives focus:
https://jamesmfriedman.github.io/rmwc/text-fields
image


arpontes:
I was reading your code and this arbitrary value in a style (marginTop 3px) really puzzled me. This wasnt matching the code style of all other components... :-)

I have just removed the line with marginTop. Thanks!

PR #314, opened 2018-09-19
Fix #259 Select will no longer ignore defaultValue

### LINKED ISSUE #259: Select does not use defaultValue
## RMWC or MDC?

Possibly RMWC

### What RMWC Version are you using?

yarn.lock:
```
rmwc@^1.7.3:
  version "1.7.3"
  resolved "https://registry.yarnpkg.com/rmwc/-/rmwc-1.7.3.tgz#9fd6785a5d5d261862b7b1540cda947f320162c6"
  dependencies:
    classnames "^2.2.5"
    hyperform "^0.9.9"
    material-components-web "0.36.1"
    mutation-observer "^1.0.3"
    prop-types "^15.6.1"
```

### What browser(s) and React Version is this bug affecting?

Firefox 61.0 and Chrome Version 67.0.3396.87 (Official Build) (64-bit)

### What build system are you using?

shadow-cljs

### What are the steps to reproduce the bug?

Add data to the item's defaultValue, inspect in both the dom and in React Dev Tools, both show:
```
Props
defaultValue: "US"
label: "Country"
onChange: fn()
options: Array[248]
placeholder: ""
```

### What is the expected behavior?

It should select that item in the options as the selected item.

### What is the actual behavior?

Shows up as blank.

### Any other information you believe would be useful?
 I've tried many different ways of getting the component reactive.  If I set the value, I can't change the displayed value, although the underlying variable does change.

Deilv:
Currently if value has been set rmwc will set defaultValue to undefined, otherwise it will set it to '', effectively meaning defaultValue is ignored.

By default React throws an error if the user defines both value and defaultValue, so rmwc shouldn't have to deal with this.

### DIFF
--- a/src/select/index.js
+++ b/src/select/index.js
@@ -138,6 +138,7 @@ export class Select extends withFoundation({
       placeholder,
       children,
       value,
+      defaultValue = '',
       outlined,
       label = '',
       options = [],
@@ -162,7 +163,7 @@ export class Select extends withFoundation({
         <SelectNativeControl
           {...rest}
           value={value}
-          defaultValue={value !== undefined ? undefined : ''}
+          defaultValue={value !== undefined ? undefined : defaultValue}
         >
           {(!!placeholder || placeholder === '') && (
             <option value="" disabled={placeholder === ''}>


jamesmfriedman:
Even though the error is annoying, RMWC does indeed have to deal with this.

Because of the way FloatingLabels are implemented, they expect the native control to have a value. If defaultValue is removed, the first value in the select is automatically chosen and the label is floated.

This problem is exacerbated by the fact that you can't set "value" without having an onChange handler. You can file an issue, but this particular PR will cause a regression in selects.

Deilv:
In that case, I will amend the PR to forcibly use defaultValue if defined, however, this check is missing in Textfield as well so you might want to look into that as well :)

jamesmfriedman:
Because of implementation details the TextField doesn't suffer from the same issues.

Deilv:
PR has been amended, this should work as well. At least it does on my test system.

jamesmfriedman:
Would still need to be defaultValue || '' to not break the select because if defaultValue is undefined we still need to set it to something.

Deilv:
You have much better understanding of this than me, but I hope this is better.

jamesmfriedman:
Sorry for the delay on this, I ended up slipping in the fix to one of my commits. I'll make sure to give you credit for finding this.

```
<SelectNativeControl
          {...rest}
          value={value}
          defaultValue={
            value !== undefined ? undefined : this.props.defaultValue || ''
          }
        >
```

PR #321, opened 2018-09-21
docs(Installation) suggest adding bg class to <body>

tstirrat:
You may have a better way but this was the only way that seemed to work for a dark theme. 

I'm happy to add a dark theme to the docs in a follow up PR.


tstirrat:
For dark themes this was still not enough..

We still need to add a `:root` variable because the `<ThemeProvider>` is not high enough in the DOM hierarchy 😞to override `--mdc-theme-background` for the `<body>` tag.

Needs:

```css
:root {
  --mdc-theme-background: #263238;
}
```

Perhaps `<ThemeProvider>` could add a `:root` entry specifically for `--mdc-theme-background` so that it works as expected ...

Your thoughts?

jamesmfriedman:
Ah, sorry, I'm going to have to get these changes over manually. the docs folder is actually a generated directory from a build so it can't be merged over.

jamesmfriedman:
Alright, I manually ported this over and made some tweaks. Sorry you don't get the contrib points, but I'll make sure to give you a shoutout in the release notes.

PR #337, opened 2018-10-08
fix/Tabs: make conditional tabs work even after re-render

martinratinaud:
This will fix issue https://github.com/jamesmfriedman/rmwc/issues/336

### DIFF
--- a/src/tabs/index.js
+++ b/src/tabs/index.js
@@ -118,13 +118,13 @@ export class TabBar extends withFoundation({
       JSON.stringify(
         React.Children.map(
           prevProps.children,
-          (child: any) => 'key' in child && child.key
+          (child: any) => 'key' in (child && child.key) || {},
         )
       ) !==
         JSON.stringify(
           React.Children.map(
             this.props.children,
-            (child: any) => 'key' in child && child.key
+            (child: any) => 'key' in (child && child.key) || {},
           )
         );
 


jamesmfriedman:
Thanks, I've known about this for a while, but haven't run back into it since I don't often do this. I'll get your PR merged in.

I appreciate you hunting it down and fixing it!

martinratinaud:
Thanks a lot @jamesmfriedman. When do you think the @rmwc/tabs package will be published on npm ?

jamesmfriedman:
When issues get fixed, I try to patch them day of. There was another minor issue with TextField classNames so I'll get that fixed as well. So sometime this morning.

There was a small error in your logic. Can you verify this fixes your issue still? 

```
(child: any) => (child && child.key) || {}
```

martinratinaud:
It does not seem to me that this is the same behavior.
I'm not sure what you wish to achieve but in my case `child.key` is always `null`.
Maybe what you want to test is if `value` have changed
in this case it would be (child: any) => (child && child.props && child.props.value) || {}

Or you want to check if any of the props have changed
in this case it would be (child: any) => (child && child.props) || {}

But for the record, I'm not sure stringifying props is the most recommended way to check for differences, we should more enumerate all changing values

jamesmfriedman:
The root of this issue is actually extremely complicated. Basically, you can't guarantee the children of the TabBar component is actually a Tab because of things like styled components, or any other component that might encapsulate the Tab.

The TabBar and Tab component will get an internal rewrite soon. In the meantime, as long as this fix works I'll keep it in there.

martinratinaud:
Ok thanks, I'll check later if package has been published, thanks again for the great work

Issue #345, opened 2018-10-21
Snackbar index.d.ts error


dave-hillier:
I'm unable to use Snackbar from typescript in 3.0.5. I get the following error. 

```
(6,29): Could not find a declaration file for module '@material/snackbar/dist/mdc.snackbar'. '/Users/daveh/source/repos/housekeeper-service/service/ClientApp/node_modules/@material/snackbar/dist/mdc.snackbar.js' implicitly has an 'any' type.
        Try `npm install @types/material__snackbar` if it exists or add a new declaration (.d.ts) file containing `declare module '@material/snackbar/dist/mdc.snackbar';`
```
As a workaround I removed line 6 @rmwc/snackbar/index.d.ts and replaced MDCSnackbar on line 120 with object.

dave-hillier:
I don't understand how the types are generated otherwise I'd try help

jamesmfriedman:
The problem is actually upstream in definitely typed. They are apparently lacking a definition for the snack bar module.

jamesmfriedman:
You can leave this open. It needs to be fixed!

dave-hillier:
One difference to note is the `syncWithProps` method. In these examples it uses PropsT 

https://github.com/jamesmfriedman/rmwc/blob/master/src/slider/index.js#L130
https://github.com/jamesmfriedman/rmwc/blob/master/src/tabs/index.js#L78
That isL `syncWithProps(nextProps: TabBarPropsT) {`

However, in snackbar it's different:
 https://github.com/jamesmfriedman/rmwc/blob/master/src/snackbar/index.js#L118
`syncWithProps(nextProps: MDCSnackbar) {`
Should it be `SnackbarPropsT`?

jamesmfriedman:
Fixed with 3.0.7

PR #348, opened 2018-10-24
use SnackbarPropsT in syncWithProps parameter

dave-hillier:
Relates to issue https://github.com/jamesmfriedman/rmwc/issues/345 

Warning: I don't know if this breaks anything. It looks reasonable to me but it's meant to be illustrative of my comment: https://github.com/jamesmfriedman/rmwc/issues/345#issuecomment-431858170

The `!!` are added to turn the type from `undefined | boolean` to just `boolean`.

Though it does build correctly.

### DIFF
--- a/src/snackbar/index.js
+++ b/src/snackbar/index.js
@@ -115,15 +115,15 @@ export class Snackbar extends withFoundation({
     this.isShowing_ = isShowing;
   }
 
-  syncWithProps(nextProps: MDCSnackbar) {
+  syncWithProps(nextProps: SnackbarPropsT) {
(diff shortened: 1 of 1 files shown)


jamesmfriedman:
Thanks for this. I'm tweaking a few internals of components today so I'll get this out for a patch shortly.

To clarify, does this fully fix your issue in #345 ?

jamesmfriedman:
Cool. I'm all about enabling cross platform development. I know windows has its own version of bash right now, but let me know if you think it would be worthwhile.

As for your fix, it's live in 3.0.7

PR #351, opened 2018-10-27
fix(TextField): Don't sync empty strings.

robcaldecott:
When using a `TextField` with validation - such as `required` - it will render as invalid immediately due to the initial `sync` of an empty string, which is not the correct behaviour. The simplest fix seems to be changing `sync` to only call `setValue` for non-empty strings.

This relates to issue 280 and 297 (which seems to be a relatively recent regression.)


jamesmfriedman:
Thanks for the @robcaldecott. This was actually a slightly different issue but has been resolved and will be deployed shortly.

Issue #277, opened 2018-07-25
Are you using RMWC for a project? What is it?


jamesmfriedman:
Someone recently asked me to send them links of sites using RMWC in the wild. According to npm, there are lots of downloads, but RMWC doesnt (and never will) phone home.

Add any publicly available links here.

scalebig:
I will share when we launch our web & native-hybrid for this fall! We are an Education Tech company - founded by key engineering and product management from Macmillan Education. Its been great to use so far 👍

jamesmfriedman:
Thanks @scalebig, looking forward to seeing it.

iddan:
We're using it for internal tools in K Health. Unfortunately they are not available publicly to inspect.

groovenectar:
Awesome job with the library, it is a pleasure to use. Adapted for use on a web presence site here: https://www.newearthva.com/

iddan:
I'm using it for Python Flame Chart python-flame-chart.netlify.com

jdeanwaite:
https://nobscrypto.com
Our company is happily using it!

danielmahon:
https://www.thefanfolio.com
The fantasy sports stock market!

Check it out, let us know what you think.

We just launched the beta of this new fantasy sports game and RMWC has been great to work with. Having previously worked with a larger but slower/buggier material design library it has been refreshing to work with something so "clean" and performant. Let the team at Google do all the hard work ;) We have other projects (web app driven kiosks and such) which may be converted over to use RMWC as well. Keep up the good work!

emyann:
Very light and simple to use, thanks for this library, I used it to build https://profilr.io.

jamesmfriedman:
Love the header on that site @emyann. Nice work!

And thanks for the props @danielmahon

matthewoden:
We're currently using this at Bayer (code and sites are internal only - can't link out).

jamesmfriedman:
@matthewoden like the Asprin? That absolutely does help :)

Dosant:
@jamesmfriedman, We've just started using it for our landing: https://www.meetter.ai/
Exactly what we've needed to keep it light, thank you!

jamesmfriedman:
Awesome @Dosant! Love the lander, nice and clean.

markrmullan:
Our company is happily using RMWC for some of our customer-facing applications at www.clover.com/dashboard. Ways to obtain login creds include:

- Becoming a business
[...]

PR #365, opened 2018-11-17
fix(Menu): Set aria-hidden dynamically for A11Y

markrmullan:
We use RMWC for some of our web applications, and it recently underwent an audit for ADA compliance. Some of our dropdown menus were flagged as improperly using the aria-hidden attribute, as it was set to true regardless of whether or not the menu was open, and contained nested elements which were subsequently ignored by screen readers.

This PR fixes aria-hidden for the Menu component.

Here's the violation in our web app:

screen shot 2018-11-16 at 2 24 14 pm

After fix:

screen shot 2018-11-18 at 4 16 35 pm
screen shot 2018-11-18 at 4 16 27 pm

### DIFF
--- a/src/menu/index.js
+++ b/src/menu/index.js
@@ -65,8 +65,7 @@ export const MenuItems = simpleTag({
   tag: List,
   classNames: 'mdc-list mdc-menu__items',
   defaultProps: {
-    role: 'menu',
-    'aria-hidden': 'true'
+    role: 'menu'
   }
 });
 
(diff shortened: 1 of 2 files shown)


jamesmfriedman:
Mark this is excellent, thank you for finding this. A few things. 

- can you update your PR to not include package-lock.json
- shouldn’t the aria label be on the menuroot and not the menu items?

Other than that it looks good. Extra cool if this library is being used for Clover, feel free to comment on #277 with your use case.

markrmullan:
Hi James, thanks for such a quick turnaround!

- Done
- Since the MenuRoot, in addition to the MenuItems wrapper, is only visible to the user when the Menu has been opened, this does make sense. Done. Screenshots updated as well.
- Commented on #277 🙂

PR #368, opened 2018-11-19
Do not override canvas polyfill if already present

### LINKED ISSUE #366: Do not override canvas in test polyfill if already present
Would it be possible to publish the jest polyfill for rmwc testing either as a separate package or included in the current rmwc package? That will make unit testing easier for people using RMWC and not needing to copy the polyfill.

https://raw.githubusercontent.com/jamesmfriedman/rmwc/master/src/base/testPolyfill.js

TIA

bartvde:
fixes #366

### DIFF
--- a/src/base/testPolyfill.js
+++ b/src/base/testPolyfill.js
@@ -25,13 +25,15 @@ export default () => {
       value: {}
     });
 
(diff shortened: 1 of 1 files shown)


bartvde:
Thanks for the quick merge.
I have a workaround, so waiting for the next release should be fine.

PR #378, opened 2018-12-07
fix(data-table): cell height of mdc-select

eranshmil:
Before:
image

After:
image


jamesmfriedman:
Thank you! Broke when the new select was released.

PR #388, opened 2019-01-03
fix(Select): use a combination of label and value as key for options

jesnil01:
Fixes: https://github.com/jamesmfriedman/rmwc/issues/387


jamesmfriedman:
Thanks!

PR #409, opened 2019-02-12
5.x Fixes typo in Collapsible List css path

danielmahon:
I'm assuming this needs to be fixed...

### DIFF
--- a/src/list/README-COLLAPSIBLE.md
+++ b/src/list/README-COLLAPSIBLE.md
@@ -5,7 +5,7 @@
 - import from **'@rmwc/list'**;
 - Import styles:
   - import **'@material/list/dist/mdc.list.css'**;
-  - import **'@material/list/collapsible-list.css'**;
+  - import **'@rmwc/list/collapsible-list.css'**;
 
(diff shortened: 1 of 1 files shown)


danielmahon:
Sorry for the world's tinniest pull-request. Let me know if you'd rather me just open an issue for anything else small like this I might find while upgrading to v5. Thanks for all your work! Hopefully, I can contribute some "larger" updates in the future.

jamesmfriedman:
Every little bit helps. Anything you find, any bugs report immediately. v5 goes live on Thursday

danielmahon:
@jamesmfriedman  I have noticed two things, havent narrowed them down quite yet, but Ill open issues if you dont already know about them.

1. Circular progress items seems to flash large black circles on page load, even after waiting for fonts to load, not sure whats going on there yet (very inconsistent, but didnt happen in v4)

2. Getting a lot of "Synthetic event" errors/warnings when opening/closing drawer, and performance drops (not sure if its something im doing yet)
```
Warning: This synthetic event is reused for performance reasons. If you're seeing this, you're accessing the property `nativeEvent` on a released/nullified synthetic event. This is set to null. If you must keep the original synthetic event around, use event.persist(). See https://fb.me/react-event-pooling for more information.```

jamesmfriedman:
I would need some repros on those. Not seeing them in my own usage

danielmahon:
No prob, I've been using React Hooks exclusively as well, which may be the cause as I may have something stuck in a loop or whatever as well. If I can't figure it out I'll post a repo and issue.

PR #410, opened 2019-02-13
5.x Adds consumeProps to CollapsibleList

danielmahon:
I was getting a `Warning: Unknown event handler property 'onOpen'. It will be ignored.` I'm assuming all those props need to be consumed for the root component?

Still not very familair with the typescript rewrite so I could be missing more...

### DIFF
--- a/src/list/collapsible-list.tsx
+++ b/src/list/collapsible-list.tsx
@@ -20,7 +20,8 @@ interface CollapsibleState {
 
 const CollapsibleRoot = componentFactory<{}>({
   displayName: 'CollapsibleRoot',
-  classNames: ['rmwc-collapsible-list']
+  classNames: ['rmwc-collapsible-list'],
+  consumeProps: ['handle','open','onOpen','onClose']
 });
 
 const possiblyFocusElement = (el: Element | null) => {


jamesmfriedman:
I ended up having to push another release out this morning and made the applicable changes. Sorry for stealing the contribution from you! Looking forward to more in the future :).

danielmahon:
@jamesmfriedman I like the CollapsibleList a lot! I had built something custom before but I'm using it now, though currently I'm using it as more of a "panel", so maybe having a variant that doesn't require using it as a list component would be nice? It works as is but I disable the actual ListItem part and change the orientation of the arrow. Maybe wrapping the handler and content box in an anchor wrapper?

feb-13-2019 13-36-03

jamesmfriedman:
I'd love to see a code snippet of how you have it implemented to better understand. The code doesn't actually require anything from Lists, but it does use them to correct focus and keyboard events.

danielmahon:
Well first... I'm an idiot. I left the `SimpleListItem` component in the handle (from the docs page) and didn't realize it until just now. So taking that out simplified it a bit as now the handle is just the `IconButton`. Couple things I noticed with it though is the `onOpen` and `onClose` arent fired by default if you remove the `SimpleListItem` from the handle, though I suppose that's to be expected. So I just had to add the `onClick` handler to the button. Also, it looks like `className` isn't getting passed down to the root `CollapsibleList` element so you can't wrap it in `styled`, etc. Prob another simple consume bug. Here's how I'm using it now (careful, rough code ahead ;)

```js
...
const ReponsiveCard = styled(Card)`
  position: relative;
  width: 100%;
  height: 100%;
  .mdc-card__actions {
    flex: none;
  }
  .rmwc-collapsible-list > .rmwc-collapsible-list__handle {
    text-align: right;
    margin-top: -48px;
    button {
      transform: rotate(0deg);
      transition: transform 0.3s;
    }
  }
  .rmwc-collapsible-list--open > .rmwc-collapsible-list__handle button {
    transform: rotate(180deg);
  }
`;
...
const CardDetails = ({ data, isNew }) => {
  const [expanded, setExpanded] = useState(false);
  const [meta, setMeta] = useState(null);
  const metaUrl = buildURL(data.src, { fm: 'json' });
  const getMetadata = async () => {
    const response = await axios.get(metaUrl);
    setMeta(response.data);
  };
  return (
    <CollapsibleList
      open={expanded}
      onOpen={() => getMetadata()}
      handle={
        <IconButton
          icon="keyboard_arrow_down"
          onClick={() => setExpanded(!expanded)}
        />
      }>
      {data.description && (
        <Text use="body2" style={{ marginBottom: '1rem' }}>
          {data.description}
        </Text>
      )}
      <CardChipSet>
        {isNew && <Chip label="new" />}
        {data.isApproved && <Chip label="approved" />}
      </CardChipSet>
      <Text use="body2" theme="textSecondaryOnBackground">
        Filename: {data.filename}
      </Text>
      <CardMeta meta={meta} />
    </CollapsibleList>
  );
};
...
```

jamesmfriedman:
Awesome. I'll see if I can get the events and className fixed for tomorrows release. If not, it'll be an easy patch.

jamesmfriedman:
FYI, cleaned up CollapsibleList code a bit more and added an example that supports what you're trying to do above to the docs (which should be live shortly, just working on publishing)

PR #431, opened 2019-03-07
fix(textfield): Add mdc-text-field-helper-line wrapper

patrickrgaffney:
Previously, the `<TextField helpText={{...}} />` rendered the `<TextFieldHelperText />` with the surrounding `mdc-text-field-helper-line` `<div>` (as per the [Material Web Components spec](https://material.io/develop/web/components/input-controls/text-field/helper-text/#html-structure)) — but you wouldn't get the wrapper `<div>` if you rendered the standalone `<TextFieldHelperText />`.

This PR move the wrapper to the standalone helper component and updates the specs to test for this. I used the `render` property from the `componentFactory()` to just add the wrapper `<div>` as that seemed simplest.

### DIFF
--- a/src/textfield/index.tsx
+++ b/src/textfield/index.tsx
@@ -298,14 +298,14 @@ export class TextField extends FoundationComponent<
       typeof helpText === 'object' && !React.isValidElement(helpText);
 
     return (
-      <div className="mdc-text-field-helper-line">
+      <React.Fragment>
         {helpText && shouldSpread ? (
           <TextFieldHelperText {...helpText as any} />
         ) : (
           <TextFieldHelperText>{helpText}</TextFieldHelperText>
         )}
         {!textarea && renderedCharacterCounter}
-      </div>
+      </React.Fragment>
     );
   }
 
@@ -503,7 +503,14 @@ export const TextFieldHelperText = componentFactory<TextFieldHelperTextProps>({
       'mdc-text-field-helper-text--validation-msg': props.validationMsg
     }
   ],
-  consumeProps: ['persistent', 'validationMsg']
+  consumeProps: ['persistent', 'validationMsg'],
+  render: (props, ref, Tag) => {
+    return (
+      <div className='mdc-text-field-helper-line'>
+        <Tag ref={ref} {...props} />
+      </div>
+    )
+  }
 });
 
 /*********************************************************************
--- a/src/textfield/textfield.spec.tsx
+++ b/src/textfield/textfield.spec.tsx
@@ -178,4 +178,22 @@ describe('TextFieldHelperText', () => {
   it('renders', () => {
     mount(<TextFieldHelperText>Hello</TextFieldHelperText>);
   });
+
+  it('sets the correct classNames', () => {
+    const el = mount(<TextFieldHelperText>Hello</TextFieldHelperText>);
+    expect(el.html().includes('mdc-text-field-helper-line')).toBe(true);
+    expect(el.html().includes('mdc-text-field-helper-text')).toBe(true);
+  })
+
+  it('uses the props', () => {
+    const el = mount(
+      <TextFieldHelperText persistent validationMsg>Hello</TextFieldHelperText>
+    );
+    expect(
+      el.html().includes('mdc-text-field-helper-text--persistent')
+    ).toBe(true);
+    expect(
+      el.html().includes('mdc-text-field-helper-text--validation-msg')
+    ).toBe(true);
+  })
 });


jamesmfriedman:
Unfortunately, this PR would break Character counts because they also have to be a child of mdc-text-field-helper-line.

Really, direct use of TextFieldHelperText should be deprecated since MDC has broken a simple way of using it. I left the component in there to not break a bunch of peoples builds.

I think the correct fix here is to just stop exporting the HelperText directly. What are your thoughts.

Side note, thanks for digging into this!

patrickrgaffney:
Ahh, I didn't even think about the character counts.

I can't speak for others but (at least since `@rmwc@5.x`) I use `TextFieldHelperText`s almost exclusively with non-`TextField` compnents: `Switch`s, `react-select`, etc. Basically just as wrapper around the `mdc` classes such that all helper text in a form matches the same styles — regardless of the actual input component.

How do you feel about leaving the `TextField` implementation as is such that the character counts continue to work, but adding that extra `mdc-text-field-helper-line` wrapper around the exported `TextFieldHelperText`. A lot of the styles for `mdc-text-field-helper-text` actually rely on it being a child of `mdc-text-field-helper-line`.

patrickrgaffney:
BTW, I work for Bayer (like the aspirin 😏) and we use `@rmwc` _a lot_ inside the company when building internal tools and products. Thanks for all your work on this!

jamesmfriedman:
The implementation can't stay as is, but the internal helpText prop doesn't have to use the exported component. It could use its own component, or just inline a div. Not ideal to have multiple implementations but that would be a fine workaround.. Do you have any examples of your usecases? I know select fields can have helpText. What other elements are you using this on?

patrickrgaffney:
I probably said that wrong. I meant the `TextField` implementation could stay as it was (what's currently on `master`), but use the `div`s directly instead of the `TextFieldHelperText` — and then the helpers implementation could have the `render()` implementation on this PR.

When I use the `TextField`, I always use the `helpText` prop:

```jsx
<TextField
  className='w-100'
  helpText={{
    persistent: true,
    validationMsg: true,
    children: 'Brief description of this value.',
  }}
  invalid={errors.description && validation}
  label='Description'
  onChange={e => updateNode(schemaPath, 'description', e.target.value)}
  outlined
  required
  rows='3'
  textarea
  value={node.description || ''}
/>
```

I use the `TextFieldHelperText` when using non-`input` form elements, for example [react-select](https://github.com/jedwatson/react-select):

```jsx
<div className='col-xs-12'>
  <label className={selectLabelClass(errors.type, validation)}>Value Type*</label>
  <Select
    className={clsx({ 'Select-error': errors.type && validation })}
    disabled={node.type !== null}
    onChange={e => updateNode(schemaPath, 'type', e)}
    options={RefData.primitiveTypes}
    placeholder='Value Type*'
    simpleValue
    value={node.type || ''}
  />
  <TextFieldHelperText
    className={clsx({ 'text-danger': errors.type && validation })}
    persistent
    validationMsg={errors.type}
  >
    Type of this value.
  </TextFieldHelperText>
</div>
```

Which looks like the following:

screen shot 2019-03-08 at 11 40 16 am

Or, even with a `@rmwc/switch`:

```jsx
<div className='col-xs-6 justify-content-center'>
  <div className='mdc-form-field param-form-switch'>
    <Switch
      checked={node.isTrue}
      disabled={node.type === 'Object' || node.type === 'Array'}
      label='Boolean Value'
      onChange={() => updateNode(schemaPath, 'isTrue', !node.isTrue)}
    />
  </div>
  <TextFieldHelperText persistent>
    Can this value ever be true?
  </TextFieldHelperText>
</div>
```

Which looks like the following:

screen shot 2019-03-08 at 11 40 30 am

----

Basically I use the `TextFieldHelperText` as a component-wrapper around the 2 `div`s with the MDC classes required to make any form element's helper text looks the same as the `TextField`s. That's actually how I noticed that it was missing the `mdc-text-field-helper-line` wrapper — the styles when using the standalone component were slightly different than using the `helpText` prop.

I can understand not want to duplicate implementations in the code, and I can easily add a similar wrapper-component to my project to accomplish the same if you'd rather remove the `TextFieldHelperText` export.

jamesmfriedman:
Gotcha! Yeah, we were talking about the same thing then. If you can update the PR accordingly that would be great. My one suggestion would be to create a new component like `InlineHelperText` that uses the same interface as `HelperText`. The reason is, anything thats created from componentFactory handles classNames, theming, that tag prop, and a bunch of other things. If you did it as an inline div, it would break passing a props object:

```jsx
// in this example, if inline divs were used internally, the className prop here would override
// the baked in className. You can add special handling for it, but thats what componentFactory does.
<TextField helpText={{className: 'bright-green', children: 'Get Help' }} />
```
On another note, Selects also have a helpText prop now you can use inline. While I don't believe MDC intended TextfieldHelpText to be used with other components, I fully support it if it works for you.

And on a last note, I'd love to add a `used by section` to the docs home page. Is there some avenue I can get permission to use Bayers logo on there? I feel like everyone else just does this without asking, but I'm trying to go the courteous route ;).

patrickrgaffney:
Awesome — that sounds perfect. I'll refactor this PR over the weekend, rebase the changes, and push back up to this PR.

Let me reach out to some people about the logo, I wouldn't anticipate any issues, but I'd rather play it safe. 😉 I'll get back to you on that one.

jamesmfriedman:
Closing stale PR

Issue #445, opened 2019-03-31
Dialog: Footer display bug and footer rendered as Object


amaesato:
RMWC v5.2.1
- Error: Footer is being rendered as an Object
<img width="765" alt="Screen Shot 2019-03-31 at 11 12 53 AM" src="https://user-images.githubusercontent.com/20821249/55292329-97ebcb80-53a6-11e9-8e01-9aa58a43465f.png">

- Bug: Footer is not being displayed if acceptLabel and cancelLabel are set to `null`

PR #446, opened 2019-03-31
fix(Dialog): Render footer in SimpleDialog

amaesato:
Issue #445 
- If `acceptLabel` and `cancelLabel` are set to null, footer will still render
- Fixed footer rendering as an Object


jamesmfriedman:
Thanks for both finding and fixing this

PR #447, opened 2019-04-04
add trailing icon example

maracuja-juice:
An example for the trailing icon wasn't in the docs. Therefore I thought as a user that this possibility is not officially supported.

This PR adds it to the docs.


maracuja-juice:
No problem. When are the docs deployed? 🤔

jamesmfriedman:
Docs are deployed with every release or manually. I have a few other things to update as well so I’ll deploy in the next few days.

PR #451, opened 2019-04-16
fix(Ripple): Fix ripple gets triggered twice on touch devices

yarnaimo:


jamesmfriedman:
Nice! Have never seen this issue before (even on touch devices). Your solution looks fine enough, however isTouched never gets set back to false. I know it wouldn’t seem like it would need to be if you are on a touch device, but hybrid devices with both touch and mouse would get messed up. Maybe on activate sets it back to false?

PR #455, opened 2019-04-22
docs(Menu): Remove duplicated SimpleMenu docs

maracuja-juice:
As pointed out by @dohomi the docs of the SimpleMenu component are duplicated. This PR only leaves one version behind. 
The duplication was introduced by the big documentation update.

### DIFF
--- a/src/menu/README.md
+++ b/src/menu/README.md
@@ -191,22 +191,3 @@ A Simplified menu component that allows you to pass a handle element and will au
 | `rootProps` | `Object` | By default, props spread to the Menu component. These will spread to the MenuSurfaceAnchor which is useful for things like overall positioning of the anchor. |
 
 
-## SimpleMenu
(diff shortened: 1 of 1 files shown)


jamesmfriedman:
Beleive it or not, those Readme files are generated. The bug is either

- Duplicates in the readme.tsx file
- Two components accidentally sharing the same displayName which would be a bug

maracuja-juice:
Didn't know that these are generated. I can't find the duplication in the readme.tsx. 🤔

dohomi:
It might be that `MenuSurface` is not described in the docs? Or does it has the same props as `Menu`?

maracuja-juice:
For me it seems like `SimpleMenuSurface` is replaced with `SimpleMenu` in the docs.

jamesmfriedman:
This was fixed yesterday with some updates to the docs. Had to do with an improper displayName being set on a component.

PR #473, opened 2019-07-23
Fix typo on installation documentation

fjorgemota:
Yeah, I know it's not the biggest PR you ever received but I observed that while reading the installation page and I throught..."well, why not?" :P


jamesmfriedman:
Thanks!

PR #477, opened 2019-08-07
support react useRef with @rwmc's inputRef

### LINKED ISSUE #475: support use of react useRef hook for inputRef on TextField
currently, `inputRef` is typed as `(ref: HTMLInputElement | HTMLTextaAreaElement | null) => void`

so today, we need to use this method to get at the input ref

```javascript
let input: HTMLInputElement | HTMLTextaAreaElement | null;

<TextField
  ...
  inputRef={(ref) => input = ref}
  ...
/>
```

however with the `useRef` hook in latest versions of react, another potentially more convenient pattern of getting access to the underlying ref could look as follows

```javascript
const input = useRef<HTMLInputElement>(null);

<TextField
 ...
 inputRef={input}
 ...
/>
```

would you be interested in this functionality?

eleith:
fixes #475 

1) update inputRef's type to allow for react's `MutableRefObject` in addition to functional ref captures for textfield, switch and checkbox components
2) added tests to ensure passing in both types are supported

### DIFF
--- a/src/checkbox/checkbox.spec.tsx
+++ b/src/checkbox/checkbox.spec.tsx
@@ -49,4 +49,19 @@ describe('Checkbox', () => {
     const el = mount(<Checkbox className={'my-custom-classname'} />);
     expect(!!~el.html().search('my-custom-classname')).toEqual(true);
   });
+
+  it('supports inputRef as an object reference', () => {
+    const inputObjectRef: any = { current: null };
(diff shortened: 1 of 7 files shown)


jamesmfriedman:
Thanks for this (and the added tests). I'll take a look at it this week and get a patch release out.

jamesmfriedman:
Thanks for this, working on some new features and bugfixes, will go out with the next release.

PR #496, opened 2019-09-28
Slider: sync min/max props before value to avoid warnings.

rjdestigter:
If `props.value` and `props.max` change simultaneously (e.g. increasing both by 1) the out of bounds warnings are still logged. I have changed the order of syncing between value and max/min to avoid that.


jamesmfriedman:
Thank you kindly. Will go out with the next patch release.

PR #537, opened 2019-12-12
Fix data table theming

dave-hillier:
Hi!

I haven't actually built and tested this change (I just made the changes to my node_modules folder), but I figured that the easiest way to explain was just to write the code. Looks like you missed the word "text" from the variables and therefore the data table does not theme correctly.

Hope this is useful!


jamesmfriedman:
Thank you! You are absolutely correct. Gotta find a way to add type safety to css variables...

Issue #533, opened 2019-12-09
Select: placeholder and value


jfily:
- What RMWC Version are you using [major.minor.patch]:

Latest : 5.7.1

- Name your build system [Webpack, Rollup...]:

Webpack

- Describe the bug with as much detail as possible:

I try to make an optional select with the placeholder prop but when I use the "value" prop, the placeholder disapear : https://codesandbox.io/s/rmwc-sandbox-zue4h

- What happened, and what was supposed to happen:

After checking the code, this is intended but I wonder why.
```javascript
{!this.props.value && !this.props.defaultValue && (
   <option value="" disabled={placeholder === ''}>
       {placeholder}
   </option>
)}
```

So is there another way to create an optional select with a placeholder ?

jamesmfriedman:
Thats definitely a bug, and some very shoddy logic! A workaround for now.

https://codesandbox.io/s/rmwc-sandbox-ijrqo

jfily:
This is what i'd do as a fix :

```typescript
{placeholder !== undefined && (
    <option value="" disabled={rest.required}>
        {placeholder}
    </option>
)}
```

jamesmfriedman:
Feel free to PR it If you’d like. Working hard on V6, will definitely get this in there.

jamesmfriedman:
Released and fixed in 5.7.2

PR #541, opened 2019-12-31
Update 5.7.2 changelog with fix for issue #533

anirban09:
Hey, it looks like the changelog for version 5.7.2 didn't include information about pull request #535, which fixed issue #533.
Realized this when we upgraded to 5.7.2 and some of our code that depended on the old behaviour broke.

So, just listed that change in the changelog too.


jamesmfriedman:
Thanks! Changelog generation depends on a specific format for commit messages, so taht one might have just been incorrect.

PR #546, opened 2020-01-12
Make z prop optional

pahan35:
As I see, `z` has default value. IMHO it should be optional. With required mode I see false positive warnings in my IDE

### DIFF
--- a/src/elevation/index.tsx
+++ b/src/elevation/index.tsx
@@ -5,7 +5,7 @@ import { componentFactory, wrapChild } from '@rmwc/base';
 /** The Elevation Component */
 export interface ElevationProps {
   /** A number from 0 - 24 for different levels of elevation */
-  z: number | string;
+  z?: number | string;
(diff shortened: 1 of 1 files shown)


jamesmfriedman:
Curious @pahan35, what benefit do you get from omitting this prop? Without it the component does nothing. Without elevation, why not just set it to zero?

pahan35:
I just notice such notification from IDE and started to investigate what is wrong here
image

As I appeared, `z` is in defaultProps, but it isn't marked as optional in the interface.

For me, it looks like a bug.

The only benefit for me is avoiding false warnings in my IDE without modifying the existing code base. And it looks logical: if the component works as expected (I hope so, I'm not an expert in styles, it was here before) without that prop why do we need it?

pahan35:
Thanks for the merge

jamesmfriedman:
Sorry it took FOREVER.  V6 was a monster to get out. Will go out in the next patch release.

PR #570, opened 2020-03-20
Feat/dialog/include on opened on closed

### LINKED ISSUE #569: bug(dialog): onStateChange and onOpen and onClose fire out of order
I am noting on 6.0.0-alpha15 that `onStateChange` does not sync as expected with `onOpen` and `onClose`

take for example (https://codesandbox.io/s/rmwc-sandbox-8ejxb)

```js
function Example() {
  const [open, setOpen] = React.useState(false);
  return (
    <React.Fragment>
      <Dialog
        open={open}
        onOpen={() => console.log("OPEN")}
        onClose={evt => {
          console.log("CLOSED");
          setOpen(false);
        }}
        onStateChange={state => console.log({ state })}
      >
        <DialogTitle>Dialog Title</DialogTitle>
        <DialogContent>This is a standard dialog.</DialogContent>
        <DialogActions>
          <DialogButton action="close">Cancel</DialogButton>
          <DialogButton action="accept" isDefaultAction>
            Sweet!
          </DialogButton>
        </DialogActions>
      </Dialog>

      <Button raised onClick={() => setOpen(true)}>
        Open standard Dialog
      </Button>
    </React.Fragment>
  );
}
```

Console output is as follows

```
OPEN 
{state: "opening"}
{state: "opened"}
CLOSED
{state: "closing"}
{state: "closed"}
```

I would expect  OPEN to come after opening and CLOSED  to come after closing. A reasonable work around would be to only use `onStateChange` when needing that sort of control. Except, `onStateChange` does not pass the action. Which seems to be the way the foundation operates.

mgr34:
closes #569 

1. adds `DialogOnOpenedEventT`  and `DialogOnClosedEventT` and formally adds methods `onOpened` and `onClosed` 
2. updates documentation to include onOpened and onClosed 
    - At least I believe that to be the case. I edited readme.tsx only. If I understand properly it will be added automatically from there
3. modifies dialog.spec.tsx to also listen for `onOpened` and `onClosed`

Left outstanding, in my opinion, is the need for `onStateChanged`. Is it necessary. At the very least should it not also include the action in close and closed instances? I suppose it would be a breaking change, but if ever was the time to implement a braking change.... ;)

### DIFF
--- a/src/dialog/dialog.spec.tsx
+++ b/src/dialog/dialog.spec.tsx
@@ -42,17 +42,21 @@ describe('Dialog', () => {
   });
 
   it('Dialog lifecycle', async done => {
+    let open = 0;
     let opened = 0;
+    let close = 0;
     let closed = 0;
     let lifecycle: string[] = [];
 
     const el = mount(
       <Dialog
-        onOpen={() => opened++}
+        onOpen={() => open++}
+        onOpened={() => opened++}
         onClose={() => {
-          closed++;
+          close++;
           el.setProps({ open: false });
         }}
+        onClosed={() => closed++}
         onStateChange={state => lifecycle.push(state)}
       >
         <DialogTitle>Dialog Title</DialogTitle>
@@ -77,7 +81,9 @@ describe('Dialog', () => {
     const cancelButton = el.find('button.mdc-dialog__button').first();
     cancelButton.simulate('click');
     await wait(250);
+    expect(open).toBe(2);
     expect(opened).toBe(2);
+    expect(close).toBe(2);
     expect(closed).toBe(2);
     expect(lifecycle).toEqual([
       'opening',
--- a/src/dialog/dialog.spec.tsx
+++ b/src/dialog/dialog.spec.tsx
@@ -46,7 +46,6 @@ describe('Dialog', () => {
     let opened = 0;
     let close = 0;
     let closed = 0;
-    let lifecycle: string[] = [];
 
     const el = mount(
       <Dialog
@@ -57,7 +56,6 @@ describe('Dialog', () => {
           el.setProps({ open: false });
         }}
         onClosed={() => closed++}
-        onStateChange={state => lifecycle.push(state)}
       >
         <DialogTitle>Dialog Title</DialogTitle>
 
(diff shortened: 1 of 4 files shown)


jamesmfriedman:
One of the reasons these props weren't added is because it leads to confusion for most people "Do I use onOpen, onOpened, or onClose, onClosed... They're too similar and for the majority of people , onOpen and onClose are the behavior that fits with all controlled components.

I was just reading the issue. Does onStateChange not work on this branch?

mgr34:
Thanks for the quick response.

[As speced](https://github.com/material-components/material-components-web/tree/master/packages/mdc-dialog#mdcdialogadapter) `onClose` and `onClosed` each emit the action. OnStateChange does not. As such, it does not satisfy my needs. Lastly, I would encourage users of mdc-react to transition to this library. That library does make use of these props. It could make converting their projects easier.

jamesmfriedman:
Fair enough, can you add one more commit to remove onStateChange then? Same commit syntax, just use breaking(Dialog): as a prefix.

mgr34:
Sure thing. If you had any hesitations you could always deprecate onStateChange and mark it for removal later on :)

jamesmfriedman:
Oh I obey SEMVER and have avoided breaking changes for an entire year with point releases. If there is going to be a breaking change, this is the time to do it.

Also, I never liked the term onStateChange...

mgr34:
agreed. 

if I recall the mdc-dialog component had a bit of a rewrite about 12-18 months ago where they added those new adapter methods. At the time I seem to recall not liking the additions, finding them trivial. But sure enough, I've used them in my projects. 

Same with `escapeKeyAction` and `scrimClickAction` . Both of which I have also managed to use. They also seem to be missing from this project. I can create a issue for you. Not sure when I could find time to implement them.

jamesmfriedman:
Yeah, I have a prop called "preventOutsideDismiss" which I just fixed a bug for to include the ESC key. But yeah, just open a feature request, both of those seem trivial to add. I can get them into this release.

jamesmfriedman:
Thanks!

PR #572, opened 2020-03-20
Feat/list/listdivider role

mgr34:
Missing `role=separator` from <ListDivider/>

see: https://github.com/material-components/material-components-web/tree/master/packages/mdc-list#list-dividers
>NOTE: the role="separator" attribute on the list divider. It is important to include this so that assistive technology can be made aware that this is a presentational element and is not meant to be included as an item in a list. Note that separator is indeed a valid role for li element

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ v6.0.0 is coming!
 
 TL;DR
 - `npm i rmwc@next` or `npm i @rmwc/button@next`.
-- Coming out sometime early spring
+- Release is close. Will be switching from alpha to RC within the next week.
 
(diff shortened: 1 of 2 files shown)


mgr34:
oh...Not sure how your update to the README.md on the master branch made it in here. I don't remember merging that. Should I resubmit?

mgr34:
I'm going to resubmit this with this change, as well as the other changes re:<Tag/> and using ul/li's instead of <div>'s

jamesmfriedman:
Oh that doesn't matter, you probably just updated / merged from master

PR #583, opened 2020-03-23
Feat/menu/misc enhancements

### LINKED ISSUE #579: bug(menu): focusOnOpen has no default value
on the 6.0.0-alpha branches The documentation for `<Menu/>` notes that `focusOnOpen` defaults to `true`.  This is not the case, it defaults to undefined. 

I'd like to submit a PR, but I am a bit confused how to set a `defaultValue`. I've mentioned this in discord if you prefer to have the discussion there. I'm only so familiar with TS, but in this case, its the handy `createComponent` that is throwing me off. 

I could change the foundation `useEffect` hook to look for props.focsOnOpen to be `undefined || true` but imagine setting a true defaultValue maybe preferred.

### LINKED ISSUE #580: feat(list): ability to wrap focus
(PR forthcoming)

prop to have list wrap focus. When reaching the end/top of the list the arrow keys should wrap focus back to the beginning/end of the list. `wrapFocus` would proxy to the foundation's `setWrapFocus()` method.

### LINKED ISSUE #581: feat(list): implement missing adapter method `isRootFocused`
(PR forthcoming)

The list adapter is missing `isRootFocused` which is utilized by the `foundation.handleKeyDown`

https://github.com/material-components/material-components-web/blob/e75deb8540fa70236087d335c9cd6280bd643285/packages/mdc-list/foundation.ts#L182

### LINKED ISSUE #582: feat(list): add vertical prop
(PR Forthcoming)

which would Proxies to the foundation's `setVerticalOrientation()` method. 

---
I don't find much use in it, but while I'm here it seems straight forward enough.

mgr34:
1. fix(menu): focusOnOpen will be treated as true when undefined (closes #579 )
2. feat(list): adds wrapFocus property (closes #580 )
3. feat(list): adds vertical property (closes #582 )
4. feat(list): implement missing adapter method isRootFocused (closes #581) 

> note: was hoping to get to a few other issues related to the menu today. Hence the branch name.

### DIFF
--- a/src/list/foundation.tsx
+++ b/src/list/foundation.tsx
@@ -247,7 +247,9 @@ export const useListFoundation = (props: ListProps & React.HTMLProps<any>) => {
   useEffect(() => {
     foundation.layout();
     foundation.setWrapFocus((props.wrapFocus || props.wrapFocus === undefined));
-    foundation.setVerticalOrientation((props.vertical || props.vertical === undefined));
+    foundation.setVerticalOrientation(
+      (props.vertical || props.vertical === undefined)
+    );
   }, [foundation, props.wrapFocus, props.vertical]);
 
   return { ...elements };
--- a/src/list/foundation.tsx
+++ b/src/list/foundation.tsx
@@ -141,7 +141,8 @@ export const useListFoundation = (props: ListProps & React.HTMLProps<any>) => {
         },
         isFocusInsideList: () => {
           return !!rootEl.ref?.contains(document.activeElement);
-        }
+        },
+        isRootFocused: () => document.activeElement === rootEl.ref,
       });
     }
   });
@@ -203,6 +204,7 @@ export const useListFoundation = (props: ListProps & React.HTMLProps<any>) => {
   const handleKeydown = useCallback(
     (evt: React.KeyboardEvent<HTMLElement> & KeyboardEvent) => {
       props.onKeyDown?.(evt);
+
 
       const index = getListItemIndex(evt);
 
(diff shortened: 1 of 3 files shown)


src/list/foundation.tsx
```diff
@@ -244,7 +246,11 @@ export const useListFoundation = (props: ListProps & React.HTMLProps<any>) => {
   // layout on mount
   useEffect(() => {
```
jamesmfriedman: I think each prop should have its own useEffect hook.

Originally, foundation.layout would run only when the foundation variable changed, which is actually never. Aka, it would only run once on mount.

I'm not actually certain what the other props actually do. It is possible you want layout to always run when they change as well, but I think you probably only want the foundation callbacks to to be called when the individual prop changes. You wouldn't want to call `setWrapFocus` when the vertical prop changed for instance.

jamesmfriedman:
Comment above, but you rock. Thank you so much already for the contributions. Making this lib better by the minute.

src/list/foundation.tsx
mgr34: Hi @jamesmfriedman. I think you are right. The `foundation.layout()` method does not need to be recalled after `vertical or `wrapFocus` is set. I am not sure that my solution is ideal, insofar as handling default values. Is it correct that at present you do not have a way to handle default values in this project?

jamesmfriedman: Surprisingly, I don't think any default values are actually needed anywhere in the library, this might be one of the only cases.

Everything else looks great

PR #585, opened 2020-03-29
fix(Icon): Resolves build-time `TypeError: Cannot read property "name" of undefined`

4cm4k1:
Howdy! I upgraded to v6.0.2 and noticed a build-time error (my build logs: https://zeit.co/4cm4k1/personal-website/m6doong9t), so I went ahead and fixed it. Let me know if more is needed for this PR. Thanks! 😄


jamesmfriedman:
Ah, thanks! Didn't trigger in my test cases. Fix one problem, cause another... Will get a patch out right now.

jamesmfriedman:
Patch released, v6.0.3

Issue #602, opened 2020-04-07
Navigation Drawer List Item Icon color doesn't use theme color when in active state, differs from the text color.


kintz09:
This is an issue with a MDC styles that will require a patch in '@rmwc/theme/theme.css'.

- _What RMWC Version are you using [major.minor.patch]_: 6.0.9

- _Name your build system [Webpack, Rollup...]_: Ejected CRA 

- _Describe the bug with as much detail as possible:_ When using a custom theme, the color of the icon in the active menu item inside the navigator drawer is incorrect. This occurs even with the theme.css stylesheet loaded.

- _What happened, and what was supposed to happen:_ The icon color was the default purple instead of the custom color as defined by the theme. I expected the theme.css stylesheet to fix it like it fixed the color of the text.

![image](https://user-images.githubusercontent.com/10118752/78624172-d16ad880-783d-11ea-9e4b-017c5266115c.png)

jamesmfriedman:
Fixed in 6.1.2.

PR #603, opened 2020-04-07
Added a hotfix for mdc style bug for issue #602

kintz09:
Added a hotfix for the incorrect nav drawer list item graphic color when activated which is caused by a bug in mwc. Let me know if you have any questions.


jamesmfriedman:
Looks good, and thank you. Will go out with the next patch release.

PR #640, opened 2020-05-22
fix(TopAppBar): Wrong className for TopAppBarFixedAdjust

### LINKED ISSUE #634: Wrong className for TopAppBarFixedAdjust
- What RMWC Version are you using [major.minor.patch]:

6.0.14

- Name your build system [Webpack, Rollup...]:

Webpack (Next.js)

- Describe the bug with as much detail as possible:

When I try to use `TopAppBarFixedAdjust` with `dense` prop. RMWC adds both `mdc-top-app-bar--fixed-adjust` and `mdc-top-app-bar--dense-fixed-adjust` to className.

``` typescript
<TopAppBarFixedAdjust dense />
```

``` html
<div class="mdc-top-app-bar--fixed-adjust mdc-top-app-bar--dense-fixed-adjust"></div>
```

But, `mdc-top-app-bar--dense-fixed-adjust` is not an option of `mdc-top-app-bar--fixed-adjust` but an independent component. So it should render like following
``` html
<div class="mdc-top-app-bar--dense-fixed-adjust"></div>
```

See the [Code Sandbox](https://codesandbox.io/s/rmwc-typescript-sandbox-cijzc) for demo.

- What happened, and what was supposed to happen:

The problem of adding both classnames happens when I use it on the mobile web. The default top app bar shrinks its height when the window size is smaller than 600px. So `mdc-top-app-bar--fixed-adjust` also shrinks its top margin.

Since (media query + class name) has higher priority than (class name only) the top margin value of `mdc-top-app-bar--fixed-adjust`, which is greater than margin value for dense one, will be used for rendering. So it breaks UI.

So `TopAppBarFixedAdjust` should add classnames exclusively.

Hardtack:
It fixes #634 

This PR adds some tests for checking class names in `TopAppBarFixedAdjust` components and updates code to pass the tests

### DIFF
--- a/src/top-app-bar/index.tsx
+++ b/src/top-app-bar/index.tsx
@@ -227,15 +227,24 @@ export interface TopAppBarFixedAdjustProps {
 export const TopAppBarFixedAdjust = createComponent<TopAppBarFixedAdjustProps>(
   function TopAppBarFixedAdjust(props, ref) {
     const { dense, denseProminent, prominent, short, ...rest } = props;
-    const className = useClassNames(props, [
-      'mdc-top-app-bar--fixed-adjust',
-      {
-        'mdc-top-app-bar--dense-fixed-adjust': props.dense,
-        'mdc-top-app-bar--prominent-fixed-adjust': props.prominent,
-        'mdc-top-app-bar--dense-prominent-fixed-adjust': props.denseProminent,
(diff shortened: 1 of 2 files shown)


jamesmfriedman:
Looks good, thanks!

Issue #613, opened 2020-04-18
Enhanced Select options are clipped inside dialog


peterkle:
- What RMWC Version are you using [major.minor.patch]: 
6.0.14

- Name your build system [Webpack, Rollup...]: 
Parcel

- Describe the bug with as much detail as possible: 
With `<Select enhanced />`, the options are clipped if the parent container is not large enough. This is especially apparent when inside the `<Dialog>` component, which may be a somewhat common use case.

- What happened, and what was supposed to happen: 
The selection options are clipped. We want to see the entire list.

Sandbox: https://codesandbox.io/s/rmwc-typescript-sandbox-mlcnm?fontsize=14&hidenavigation=1&theme=dark

sangdth:
I think the usage of very long list is against the Dialog design principle, according to [Material design documentation.](https://material.io/components/dialogs#behavior) The Dialog body, IMHO, acts like a container, and in this case, both `Dialog` and `Select` works as expected, not a bug. You just use it in wrong way.

nicholaschiang:
Yeah @peterkle you should follow [the documentation's instructions](https://rmwc.io/menus) to render your `Select`'s `MenuSurface` through a portal:
```jsx
  // Somewhere at the top level of your app
  // Render the RMWC Portal
  // You only have to do this once
  import React from 'react';
  import { Portal } from '@rmwc/base';

  export default function App() {
    return (
      <div>
        ...
        <Portal />
      </div>
    )
  }
```
And then render your `MenuSurface` through the portal:
```jsx
function Example() {
  const [renderToPortal, setRenderToPortal] = React.useState(true);
  const options = ['Cookies', 'Pizza', 'Icecream'];
  return (
    <>
      <div
        style={{
          marginBottom: '10rem',
          height: '3.5rem',
          overflow: 'hidden'
        }}
      >
        <MenuSurfaceAnchor>
          <Button raised>Open Menu</Button>
          <Menu open renderToPortal={renderToPortal}>
            {options.map(o => (
              <MenuItem key={o}>{o}</MenuItem>
            ))}
          </Menu>
        </MenuSurfaceAnchor>
      </div>
      <Checkbox
        checked={renderToPortal}
        onChange={evt => setRenderToPortal(evt.currentTarget.checked)}
        label="renderToPortal"
      />
    </>
  );
}
```
Does that help @peterkle?

[...]

Issue #627, opened 2020-05-09
SimpleMenu breaks when adding "renderToPortal" prop


anthonysapien:
Here's a sandbox:
https://codesandbox.io/s/rmwc-sandbox-ybu69?file=/index.js

- What RMWC Version are you using [major.minor.patch]:
6.1.0

- Name your build system [Webpack, Rollup...]:
create-react-app

- Describe the bug with as much detail as possible:
Adding the renderToPortal prop to SimpleMenu causes the menu to no longer handle menu item selection.

- What happened, and what was supposed to happen:
When I add the renderToPortal property, if I open the SimpleMenu and click a menu item, the menu stays open and the onSelect event handler does not fire.

I expect the menu to close and the onSelect event handle to fire.

Thanks!

anthonysapien:
I just tested the standard Menu, and it has the same issue. The renderToPortal Menu does not handle selection here: https://rmwc.io/menus

mockingbird87:
Can confirm, I'm running into this issue as well!!

See my response[ here](https://github.com/jamesmfriedman/rmwc/issues/613#issuecomment-626207976) for a workaround.

jamesmfriedman:
Confirming this is fixed in 6.1.2. You can try adding it in the docs to see. https://rmwc.io/menus

anthonysapien:
Thanks for looking into it. I'm still seeing the issue on the RMWC site:
https://share.getcloudapp.com/4guj7d8N

The problem is that selecting an item doesn't close the menu.

jamesmfriedman:
That specific example the open prop is always set to true. Just add `renderToPortal` to any others, it works as expected

anthonysapien:
Apologies if I'm missing something. Here's a sandbox where I still don't see the menu close or the onSelect handler fire when I select an item:
https://codesandbox.io/s/rmwc-sandbox-fskbn

jamesmfriedman:
@anthonysapien thanks for reporting that. I think it has to do with event bubbling through React portals and the fact that MDC doesn't use react events, but standard DOM ones. I'll have to inspect more.

For now, here is a non-ideal, but working solution. https://codesandbox.io/s/rmwc-sandbox-omw4n

anthonysapien:
Cool @jamesmfriedman. Thanks for looking into it and for the workaround.

rayjd:
[...]

PR #649, opened 2020-06-01
fix(Select): fixed menu props not forwarding for enhanced select

grigory-bogush:
This bug did not allow MenuProps like "renderToPortal" to be forwarded to menu for enhanced select when passing in an object in "enhanced" prop.

### DIFF
--- a/src/select/select/index.tsx
+++ b/src/select/select/index.tsx
@@ -336,6 +336,8 @@ export const Select: RMWC.ComponentType<
     }
   ]);
 
+  const enhancedMenuProps = typeof enhanced === 'object' ? enhanced : {};
+
   const defaultValue =
     value !== undefined ? undefined : props.defaultValue || '';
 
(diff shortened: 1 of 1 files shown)


nicholaschiang:
Hey, I just realized that this is the same as [the PR I just opened](https://github.com/jamesmfriedman/rmwc/pull/655).

This should fix #613 and help with #627 too.

grigory-bogush:
> Can I just say how much I appreciate someone actually helping fix something rather than just saying it’s broken!? Approving, would be great to add a test to prevent a regression if you have time.

Sorry for not responding. I think the email got lost somewhere. I am not sure about how these tests are done, I am relatively new to the area and open source in general, but I'll try to look into it.

jamesmfriedman:
@nicholaschiang sorry for the delay on yours. I just checked it out, yours would cause a regression since we still need the {...rest} props as well. Going to merge this one in favor of that.

PR #661, opened 2020-06-29
fix(Tab): activeTabIndex always updating automatically

rbare1:
Before: 

After: 

https://github.com/jamesmfriedman/rmwc/issues/606


jamesmfriedman:
Seems reasonable. Thanks for the gif. I assume that the "Always 1" tab is the thing you were fixing?

rbare1:
@jamesmfriedman Yeah, the "Always 1" tab was the bit being fixed.

jamesmfriedman:
Thanks for doing this, will go out in the next release.

PR #662, opened 2020-07-02
Fix typings for rmwc package

bjarkehs:
Right now the typings for e.g. `import { Button } from 'rmwc'` are not imported correctly.


jamesmfriedman:
Thanks for fixing this

bjarkehs:
@jamesmfriedman Is it possible to release this?

PR #667, opened 2020-07-08
deps: remove not used deps from externals

pahan35:
It looks like there are 2 deprecated package entries that aren't used anymore.

Is it safe to delete them?


jamesmfriedman:
It should be safe to delete those, but I haven't used the lib build in a long time. Is that something you're actively using?

pahan35:
Thanks for proving my suggestion. 

I was clearing my package.json from `@material/x` packages (other team-mate added them for some reason a long ago). I've deleted everything except those two deps.

So, this PR is kind of 'Is it safe to delete them in my current project which uses rmwc@5.7.2?'

But when I tried to search those entries in the source code, I found only those entries. I've decided to make this PR to make sure it's safe to delete them and not to confuse other devs when they face a similar question.

jamesmfriedman:
Yep, that whole file can probably go but this is safe for now.

PR #670, opened 2020-07-16
Fix clicking items in Menus and enhanced Select menus

MiK546:
Currently Menus and enhanced Select menus using renderToPortal to render the pop-up containing the menu items to a Portal are broken. Clicking menu items does nothing and the onSelect callback is never called.

This seemingly happens because the useFoundation() hook from @rmwc/base seems to set null as the apiRef/listApi reference in the Menu foundation when the cleanup function of the useEffect() hook in useFoundation() is called. The same useEffect() hook then never re-sets the listApi leaving it as null. This seems to only happen when renderToPortal is used for some reason. 

As a solution the useEffect() hook now sets the apiRef of components every time it is run in addition to the apiRef being set in the useMemo() hook. This seems to fix the behaviours of both type of menus. Does this seem sensible?

I also modified the Menu story that uses renderToPortal to show the active index and added a new Select story that uses a Portal. These helped me in debugging.

This PR should fix the problems described by me and some others in https://github.com/jamesmfriedman/rmwc/issues/613.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "rmwc-main",
-	"version": "6.0.0",
+	"version": "6.1.1",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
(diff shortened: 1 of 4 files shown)


jamesmfriedman:
Looks good, and good catch. Thanks!

Issue #557, opened 2020-02-25
Allow `focusOnActivate` to be set on tabs


rbare1:
We're migrating from MDC React to RMWC, and one prop we're missing on tabs is the ability to set `focusOnActivate` which causes the active tab to steal focus on page load.

Here is the issue in MDC React where it was discussed and made configurable: https://github.com/material-components/material-components-web-react/issues/676

PR #677, opened 2020-07-27
feat(Tab): Add ability to set focusOnActivate

rbare1:
This PR allows the tabs to be configured as to whether they should automatically focus when activated. This is set to true by default in MDC Web, but can be configured through the foundation.

When true, the active tab will immediately gain focus on page load. In this case it's the "Pizza" tab in "Always 1"
chrome_2020-07-27_15-24-28

If we set the Tabs in the "Always 1" example to have `focusOnActivate={false}, then the focus will not be set on this set of tabs and focus will remain on the previous story that still has the default behavior.
chrome_2020-07-27_15-26-00

Resolves: #557


jamesmfriedman:
Looks good, thanks!

PR #688, opened 2020-08-18
Simple dependent select crash fix

grigory-bogush:
Fixes https://github.com/jamesmfriedman/rmwc/issues/686 crash, but no idea if the root cause is on `rmwc` side or `mdn`

I also added a story to demonstrate the fix, since I am not that good with testing, but I assume that would preferred..? 
Is it okay to include all of that hook stuff in tests? I haven't seen tests for such situations in the repo select.spec file.

### DIFF
--- a/src/select/select.story.tsx
+++ b/src/select/select.story.tsx
@@ -40,6 +40,48 @@ function MutatingSelect(props: any) {
   );
 }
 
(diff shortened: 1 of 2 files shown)


jamesmfriedman:
I wish typescript would consider every array index access unsafe. It's about the only source of runtime issues I hit frequently.

PR #697, opened 2020-09-17
Do not pass "wrapFocus" List prop to Tag

gabehayes:
Allowing `wrapFocus` to remain in the `...rest` parameter will throw a react-dom warning about an unrecognized prop.


jamesmfriedman:
Thanks!

PR #707, opened 2020-11-27
chore(Icon): Improve typings of icon render method

thdk:
A small PR to start my contribution for RMWC.

Minor change is that the custom render method of Icon component may no longer return a string or a number (`ReactText` type is now excluded from the `ReactNode` union type). However, that should not change compatiblity since the code would have crashed anyway if users would return a `string | number` from the custom render method. 

Now it will simply return null when the render method doesn't return a valid `ReactElement`.


jamesmfriedman (approved):
Thanks!

PR #720, opened 2021-01-29
fix dev dependency: prettier#master => prettier#main

gugu:
Prettier have update their report from master branch to main


gugu:
btw do we still need to use git dependency? I bet they released a new version since that change

PR #721, opened 2021-01-29
use prettier#main

gugu:


gugu:
btw do we still need to use git dependency? I bet they released a new version since that change

gugu:
@jamesmfriedman without this PR development version does not start

amadsen (commented):
@jamesmfriedman It looks like the github dependency on prettier was added in November 2019 - possibly to pick up what were then unpublished changes. I'm considering approving this PR to allow the dev version to build, then opening a new PR to use the latest version published on npm. Thoughts?

jamesmfriedman:
This is resolved by #740, going to close this in favor of that.

PR #727, opened 2021-03-30
fix(snackbar): support SVGs as dismiss icon

nicholaschiang:
Allows users to specify their own SVGs or images to use as the `Snackbar` dismiss icon. Mirrors the existing behavior of the `Icon` and `IconButton` components.


jamesmfriedman:
Cool, thanks!

PR #741, opened 2021-07-02
fix(Portal): Prevent children from being mounted twice

thammi:
Because `PortalChild` uses an effect and falls through to directly rendering, child components are mounted and rendered twice (once at the original position in the tree and at the portal immediately after that). As the child component is placed at a completely different place in the tree a completely new life-cycle seems to be started without reusing memos and effects. This lead to an HTTP request being sent twice in my application because an effect was started twice.

The effect was replaced with a memo to use the portal in the first render. Checking whether `document` exists at the start of the memo to keep support for server side rendering.

I hope this is the right approach to fix the issue.


panmau:
Seems good to me from a first look.
I think it would be good if you could create a small demo that demonstrates the bug.

thammi:
I created a short example using a dialog. When rendering to the portal (enable/disable using the checkbox) the counter gets incremented twice by an effect that should only be executed once during the lifecycle of the component:

https://codesandbox.io/s/rmwc-portal-bug-vldqt?file=/src/index.tsx

jamesmfriedman (approved):

EmiBemi:
Hi @thammi, this PR actually resulted in a bug with menus not being rendered to the portal. Reverting this PR helps. Could it be possible that we somehow find a different fix together? Or potentially look into why the menu is not being rendered in the correct position with the changes from this PR.

PR #746, opened 2021-12-14
fix(test): Fixed failing tests

EmiBemi:
Fixed the remaining failing tests. Use react testing library instead of enzyme for the tests that were failing.

### DIFF
--- a/src/doc-utils.tsx
+++ b/src/doc-utils.tsx
@@ -467,12 +467,21 @@ export function DocsMarkdown({ fileSrc }: { fileSrc: string }) {
                 </Tag>
               );
             },
-            paragraph: ({ children }: { children: any }) => (
+            p: ({ children }: { children: any }) => (
               <p className="docs-p">{children}</p>
             ),
-            code: ({ value }: { value: string }) => (
-              <DocsExampleBase code={value} codeOnly />
-            )
+            code: ({
+              children,
+              inline
+            }: {
+              children: any;
+              inline: boolean;
+            }) => {
+              if (inline) {
+                return <code>{children}</code>;
+              }
+              return <DocsExampleBase code={children[0]} codeOnly />;
+            }
           } as any
         }
       />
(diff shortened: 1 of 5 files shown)


src/doc-utils.tsx
```diff
@@ -467,12 +467,21 @@ export function DocsMarkdown({ fileSrc }: { fileSrc: string }) {
                 </Tag>
               );
```
thaestrup, line 468: I do not know what we use the extra id for in the heading but I think that we also need to change the 'heading' to the 'h1-h5' tags as they have changed aswell.
See
https://github.com/remarkjs/react-markdown/blob/main/changelog.md#change-renderers-to-components
https://github.com/remarkjs/react-markdown#appendix-b-components

thaestrup (changes requested):

src/doc-utils.tsx
EmiBemi: I just tried to comment out the id. Seemingly, it doesn't change anything visually. And if we no longer need the id, we no longer need to specify any markdown for h1-h6, thus we can remove it all together.

b-d-m-p:
@panmau Maybe you could have a look at this? Cheers.

b-d-m-p:
@EmiBemi In @jamesmfriedman 's version of the PR he says:

> To test this PR:
> Setup:
> 
> git pull jamesmfriedman/dep-updates
> npm i
> npm run pretest
> Things to validate
> 
> Run the docs: npm start
> Run storybook: npm run storybook
> Run the test suite: npm test
> Run a build: npm run build

Pretest was fine for me, but I am getting invalid hooks calls when running the docs. 

What version of node are you on? I loaded 16.1.0 since this PR supports 16.

JacquelineFriedberg:
Note:
On `npm run storybook` the `favorite` google icon fails, but it works on `npm start`

`npm run storybook`:
Screen Shot 2021-12-14 at 8 43 39 PM
Screen Shot 2021-12-14 at 8 42 52 PM

`npm start`:
Screen Shot 2021-12-14 at 10 09 52 PM


Otherwise tests pass. ✨

jamesmfriedman (approved):

PR #747, opened 2021-12-31
Remove findDOMNode

JohnGu9:
1) remove all React.findDOMNode
2) no more export 'version' (replace it with RMWC_VERSION. Under react-script v5 or webpack v5, modules not be allowed to export value name as 'version')


thaestrup:
Hi. Nice with a little cleaning.
We generate the changelog based on the git commit log and for that to work properly it needs to be formatted in a special way.
https://github.com/jamesmfriedman/rmwc/blob/master/CONTRIBUTING.md#submitting-pull-requests 
Just for clarity renaming version to RMWC_VERSION is a breaking change so we need to mark it as such.

b-d-m-p:
@JohnGu9 we are sorry for the delay. Are these changes still relevant? Can you update the branch with the latest commits? We promise to act quick after that.

b-d-m-p:
@jensborch @EmiBemi @thaestrup I think this is workable. I will need to solve the merge conflicts, but my greater question is _if_ we should do it because it would be a breaking change.  What do you all think?

jensborch:
@b-d-m-p I think we should resolve the conflict and get this merged. The RMWC_VERSION breaking change in src/rmwc/index.tsx should be reverted, but I think we should include it in the RMW 14 release. I will create tasks on our Kanban board

b-d-m-p:
We decided to make a new PR to the `next` branch instead of this branch, but after fixing merge conflicts the project doesn't run, which it turns out is due to issues in the `next` branch, so we need to wait till those are solved before dealing with this PR. I am leaving this PR here as a reminder for the time being and I will continue conversations about it here. You can see comments on the previous merge attempts here: https://github.com/rmwc/rmwc/pull/935.

PR #780, opened 2022-04-21
build(deps): CRA-v5, webpack-v5, router-v6, react-live-v3

thaestrup:
Multiple dependencies updated.
React-scripts v5
Webpack v5
React-router v6
React-live v3


EmiBemi:
Coverage seemed to have dropped significantly. Did all tests pass?

thaestrup:
Yes the coverage is odd. I will look into it. All the tests seems to be running though.

thaestrup:
I found the issue. The dist folder is being included when running the tests and the coverage is reported on those instead of the src files. I need to find a way to exclude dist from resolution.
It's yet again something webpack related so I will put this PR in draft until I have resolved the issue.

src/base/utils/use-knob.tsx
```diff
@@ -3,7 +3,7 @@ import { useState } from 'react';
 import * as knobTypes from '@storybook/addon-knobs';
 
 // @ts-ignore
-import { manager } from '@storybook/addon-knobs/dist/esm/registerKnobs';
+import { manager } from '@storybook/addon-knobs/dist/registerKnobs';
```
EmiBemi, line 6: Could the issue somehow be related to importing stuff from dist?

jamesmfriedman (dismissed):
This looks good and appears to be working well for me locally. The only thing I'd like to see cleaned up is the lint output. There are loads of warnings in the console. But otherwise epic job getting all of the build deps updated!

jamesmfriedman:
@thaestrup finally got his working! What a pain in the butt. Ultimately it had to do with the dist files not being cleaned out. `npm prepare` appears to be running which builds the code, but it should have also been running npm clean.

In any case, fixed that issue, and added React 18 to the test matrix. Also pointed this at next and got it updated and resolved and conflicts. I'm going to merge this, then we can a release out (finally).

PR #782, opened 2022-04-25
Mcw/v8

EmiBemi:
All components have been upgraded to version 8, except the slider component.

The slider component has been completely rewritten as of MCW 8, so we figured it would be best to save this effort of rewriting it to version 13.


jamesmfriedman (approved):
Looks good. There is an issue with the select component in safari but I'm not certain if it's RMWC or an issue from material. Basically the select isn't opening properly on desktop safari.

Other than that looks good, tests are passing and everything else is working. I'd say good enough if we're just doing this as a pre-release. If we want to to the standard release channel we might want to look at the safari select issue.

EmiBemi:
@jamesmfriedman Yikes. It turns out that the Select doesn't work on the current RMWC release (7.0.2). I suspect it's due to the custom styling in select.css. The positioning in the DOM has changed since MCW v5. If you click on what is marked on the picture below, the select should appear. I will fix this on a separate branch.
Screenshot 2022-05-14 at 16 34 54

PR #784, opened 2022-05-14
Fix select on safari

EmiBemi:


EmiBemi:
Hmm, seems like I somehow managed to format some md files in the process.

jamesmfriedman (approved):
Awesome thanks!

Issue #781, opened 2022-04-24
`preventOutsideDismiss` does not work correctly if toggled after dialog opens


L-u-k-e:
- What RMWC Version are you using [major.minor.patch]:
7.0.2

- Name your build system [Webpack, Rollup...]:
Next.js

- Describe the bug with as much detail as possible:

The `Dialog` component's `preventOutsideDismiss` prop is supposed to prevent the dialog from closing when the scrim is clicked OR when the escape key is pressed. It works for the most part, but if the prop is initially false, and then later is toggled to true *while the dialog is open*, the dialog will still close when the escape key is pressed. (Clicking the scrim does correctly prevent a dismiss, but hitting escape does not.)

Code Sandbox: https://codesandbox.io/s/rmwc-sandbox-forked-d6nv14?file=/index.js

🕵️ Investigating the cause of the issue.... in `Dialog/foundation.tsx`.

It looks like the event listener is correctly being recreated when the prop updates: https://github.com/jamesmfriedman/rmwc/blob/a59d93fe568b3a5d30806d99e23553e30b502a32/src/dialog/foundation.tsx#L104

```
  const handleDocumentKeydown = useMemo(() => {
    const bound = foundation.handleDocumentKeydown.bind(foundation);
    // Wrap the callback in a function that can
    // short circuit the escape key if we are preventing outside dismiss
    return (evt: KeyboardEvent) => {
      if (evt.which === 27 && props.preventOutsideDismiss) return;
      return bound(evt);
    };
  }, [foundation, props.preventOutsideDismiss]);
```
👍 

but I believe the code causing the issue is here, a bit further down: https://github.com/jamesmfriedman/rmwc/blob/a59d93fe568b3a5d30806d99e23553e30b502a32/src/dialog/foundation.tsx#L153

```
  useEffect(() => {
    if (props.open) {
      if (!foundation.isOpen()) {
        document.addEventListener('keydown', handleDocumentKeydown);
        foundation.open();
      }
    } else {
      if (foundation.isOpen()) {
        document.removeEventListener('keydown', handleDocumentKeydown);
        foundation.close();
      }
    }
  }, [props.open, foundation, handleDocumentKeydown]);
```

The effect is re-running when the `handleDocumentKeydown` function is recreated, BUT the new event listener is not being attached to the DOM because the open prop matches the dialog's open state.

[...]

Issue #777, opened 2022-03-30
TextField `characterCount` does not work.


jmasukawa:
Re-opening per "issue bankruptcy / full reset", this one is still an issue.

- What RMWC Version are you using [major.minor.patch]: 7.0.2

- Describe the bug with as much detail as possible:
See @nicholaschiang's original issue (https://github.com/jamesmfriedman/rmwc/issues/671), which has a lot of details about the problem, and is still an accurate description of the bug. 

This was originally reported on v6.1.3 (and I experienced on v6.1.4 and v7.0.2).

This issue is also reproducible on the RMWC demo site ("Textareas", https://rmwc.io/text-fields):
![image](https://user-images.githubusercontent.com/1621643/160910122-f0bb00cd-a5d6-46ee-a8b5-9353efcc995f.png)

I could make a code sandbox for you if that really helps, but I figured the RMWC demo site may be a good enough substitute, let me know.

Thanks for all your hard work on this project!

b-d-m-p:
@jmasukawa Thanks! I'll look into it!

EmiBemi:
Closed as it has now been fixed and released with v8!

jmasukawa:
> Closed as it has now been fixed and released with v8!

Thank you so much!!

PR #786, opened 2022-05-25
Bugfixes

jamesmfriedman:
Corrects the currently open bugs and issues in V8


src/base/portal.tsx
```diff
@@ -1,12 +1,12 @@
...
+import React, { useRef, useMemo } from 'react';
 import ReactDOM from 'react-dom';
 
 const PORTAL_ID = 'rmwcPortal';
 
 export type PortalPropT = Element | string | boolean | undefined | null;
 
 export function Portal() {
-  const el = useRef(document.createElement('div'));
+  const el = useRef<HTMLDivElement | null>(null);
```
jamesmfriedman, line 9: This was probably being done for typesafety so that el was initialized to a div, but this really didn't make any practical sense. So initializing this to null.

src/dialog/foundation.tsx
```diff
@@ -106,7 +106,7 @@ export const useDialogFoundation = (
     // Wrap the callback in a function that can
     // short circuit the escape key if we are preventing outside dismiss
     return (evt: KeyboardEvent) => {
-      if (evt.which === 27 && props.preventOutsideDismiss) return;
+      if (evt.key === 'Escape' && props.preventOutsideDismiss) return;
```
jamesmfriedman, line 109: Updating deprecated DOM api

src/dialog/foundation.tsx
```diff
@@ -162,6 +162,9 @@ export const useDialogFoundation = (
         foundation.close();
       }
     }
+    return () => {
+      document.removeEventListener('keydown', handleDocumentKeydown);
+    };
```
jamesmfriedman, line 167: This corrects issue #781. When the props change, we need to unbind the existing handlers.

src/rmwc/docs/styles/docs.css
```diff
@@ -573,9 +573,7 @@ ul li {
 
 /** Page textfields */
 #page-text-fields .live-preview .mdc-text-field {
-  margin-top: 1.5rem;
-  float: left;
-  clear: left;
+  margin-top: 1rem;
```
jamesmfriedman, line 576: Since the DOM structure changed of the TextField elements again, it broke the docs.

src/select/select/index.tsx
```diff
@@ -58,8 +58,6 @@ export interface SelectProps {
   icon?: RMWC.IconPropT;
   /** Advanced: A reference to the MDCFoundation. */
   foundationRef?: React.Ref<MDCSelectFoundation>;
-  /** Makes Select fullwidth */
-  fullwidth?: boolean;
```
jamesmfriedman, line 62: No longer supported by MDC

src/select/select/index.tsx
```diff
@@ -200,7 +198,6 @@ interface EnhancedMenuProps extends MenuProps {
   value?: string;
   defaultValue?: any;
   children?: React.ReactNode;
-  fullwidth?: boolean;
```
jamesmfriedman, line 203: No longer supported by MDC

src/select/select/index.tsx
```diff
@@ -216,12 +213,7 @@ function EnhancedMenu(props: EnhancedMenuProps & SelectHTMLProps) {
 
   let currentIndex = 0;
 
-  const className = useClassNames(props, [
-    'mdc-select__menu',
-    {
-      'mdc-menu-surface--fullwidth': !!props.fullwidth
```
jamesmfriedman, line 222: No longer supported by MDC

src/select/select/index.tsx
```diff
@@ -311,7 +303,6 @@ export const Select: RMWC.ComponentType<
     inputRef,
     helpText,
     foundationRef,
-    fullwidth,
```
jamesmfriedman, line 314: No longer supported by MDC

src/select/select/index.tsx
```diff
@@ -350,8 +341,7 @@ export const Select: RMWC.ComponentType<
       'mdc-select--required': !!props.required,
       'mdc-select--invalid': !!invalid,
       'mdc-select--with-leading-icon': !!icon,
-      'mdc-select--no-label': !label,
-      'mdc-select--fullwidth': !!fullwidth
```
jamesmfriedman, line 354: No longer supported by MDC

src/select/select/index.tsx
```diff
@@ -456,7 +446,6 @@ export const Select: RMWC.ComponentType<
             selectedIndex={selectedIndex}
             menuApiRef={setMenu}
             children={children}
-            fullwidth={fullwidth}
```
jamesmfriedman, line 459: No longer supported by MDC

src/textfield/README.md
```diff
@@ -57,7 +53,6 @@ You can make the TextField a textarea. Make sure to include `outlined` for prope
 <TextField
   textarea
   outlined
-  fullwidth
```
jamesmfriedman, line 60: No longer supported by MDC

src/textfield/textfield-foundation.tsx
```diff
@@ -32,6 +33,11 @@ export const useTextFieldFoundation = (props: TextFieldProps) => {
   const setFloatingLabel = (api: FloatingLabelApi | null) =>
     (floatingLabel.current = api);
 
+  const { content: characterCountContent } =
+    useTextFieldCharacterCountFoundation({
+      apiRef: props.characterCount ? setCharacterCounter : undefined
+    });
+
```
jamesmfriedman, line 40: #777  This was incredibly annoying to diagnose and fix. The breakdown:
- The way RMWC is setup, the foundations are initialized when components render. This is fine 99% of the time.
- In this particular case, the TextfieldFoundation requires the TextfieldCharacterCount to be initialized. This creates a really annoying race condition because the character count component will be mounted after the text field.
- Solution, get rid of the current internal TextfieldCharacterCount component and instead just initialize the foundation inside of useTextfieldFoundation and just export the actual character count number
- Extra annoying, hooks cannot be initialized conditionally, so we pass in undefined for the api ref if character count isn't in play which effectively short circuits the character count.
- And it works 🤦‍♂️

thaestrup (approved):

src/textfield/index.tsx
```diff
@@ -192,7 +187,9 @@ export const TextField: RMWC.ComponentType<
   ) : null;
 
   const renderedCharacterCounter = characterCount ? (
-    <TextFieldCharacterCount apiRef={setCharacterCounter} />
+    <div className="mdc-text-field-character-counter">
```
EmiBemi, line 190: Codecov tells us that this line is not covered by tests

PR #805, opened 2022-06-19
Rtl rewrite

EmiBemi:


src/drawer/drawer.spec.tsx
```diff
@@ -49,11 +51,12 @@ describe('Drawer', () => {
...
     );
+    expect(asFragment()).toMatchSnapshot();
   });
 
-  it('can open', (done) => {
-    const el = mount(
-      <Drawer dismissible>
+  it('can open', () => {
+    const { container } = render(
+      <Drawer dismissible open>
```
thaestrup, line 59: We could remove the "open" attribute and then rerender the drawer with open and verify.

src/snackbar/snackbar.spec.tsx
```diff
@@ -123,17 +118,14 @@ describe('SnackbarQueue', () => {
...
     });
 
-    await wait(500);
-
-    expect(el.html().includes('myNotificationTitle1')).toBe(true);
-    expect(el.html().includes('myNotificationBody1')).toBe(true);
+    waitFor(() => {
+      expect(screen.getByText('myNotificationTitle1')).toBeInTheDocument();
+      expect(screen.getByText('myNotificationBody1')).toBeInTheDocument();
+    });
```
thaestrup, line 124: What about the second notification. No verify on that?

src/snackbar/snackbar.spec.tsx
```diff
@@ -20,92 +20,87 @@ describe('Snackbar', () => {
...
   it('notifies', async () => {
     const queue = createSnackbarQueue();
-    const el = mount(<SnackbarQueue messages={queue.messages} />);
+    render(<SnackbarQueue messages={queue.messages} />);
     // check multiple notifications
     queue.notify({
       title: 'myNotificationTitle1',
       body: 'myNotificationBody1',
-      timeout: 500,
+      // timeout: 500,
```
thaestrup, line 96:
```suggestion
```

src/snackbar/snackbar.spec.tsx
```diff
@@ -20,92 +20,87 @@ describe('Snackbar', () => {
...
       onClose: () => {}
     });
 
     queue.notify({
       title: 'myNotificationTitle2',
       body: 'myNotificationBody2',
-      timeout: 500,
+      // timeout: 500,
```
thaestrup, line 103:
```suggestion
```

src/tabs/tabs.spec.tsx
```diff
@@ -138,21 +113,16 @@ describe('Tabs', () => {
 
     window.requestAnimationFrame(() => {
```
thaestrup, line 114: You are right to remove the done because that is no longer supported. What we should do instead is to make the async call into a promise and then return that promise in the test function. That way the test will wait for that promise to resolve before evaluating the expect.
That being said I have not tried that with the requestAnimationFrame before so I have no idea how jsdom is working with it.

src/top-app-bar/top-app-bar.spec.tsx
```diff
@@ -31,118 +31,127 @@ describe('TopAppBar', () => {
...
-      <SimpleTopAppBar title="TestTitle" navigationIcon={{ icon: 'foo' }} />
-    );
     // has title
-    expect(el.html().includes('TestTitle')).toBe(true);
+    expect(screen.getByText('TestTitle')).toBeInTheDocument();
 
     // has the navigation icon
-    expect(el.html().includes('mdc-top-app-bar__navigation-icon')).toBe(true);
+    expect(screen.getByText('foo')).toBeInTheDocument();
+    // expect(el.html().includes('mdc-top-app-bar__navigation-icon')).toBe(true);
```
thaestrup, line 106:
```suggestion
```

src/top-app-bar/top-app-bar.spec.tsx
```diff
@@ -31,118 +31,127 @@ describe('TopAppBar', () => {
...
     // has content
-    expect(el.html().includes('TestStartContent')).toBe(true);
-    expect(el.html().includes('TestEndContent')).toBe(true);
+    expect(screen.getByText('TestStartContent')).toBeInTheDocument();
+    expect(screen.getByText('TestEndContent')).toBeInTheDocument();
 
     // has action item
-    expect(el.html().includes('star_outline')).toBe(true);
+    expect(screen.getByText('star_outline')).toBeInTheDocument();
+    // expect(el.html().includes('star_outline')).toBe(true);
```
thaestrup, line 114:
```suggestion
```

thaestrup (changes requested):
There are a couple of things that we should change/fix but great work.

src/snackbar/snackbar.spec.tsx
EmiBemi: It's right below

src/dialog/dialog.spec.tsx
```diff
@@ -25,73 +24,32 @@ describe('Dialog', () => {
...
-    mount(
+    const { asFragment } = render(
       <SimpleDialog title="This is a simple dialog" open onClose={(evt) => {}}>
         Hello
       </SimpleDialog>
     );
+    expect(asFragment()).toMatchSnapshot();
   });
 
-  // SKIP broken
```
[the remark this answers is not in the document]
EmiBemi: Sure thing. Do you want the honor? 🥺

jamesmfriedman (dismissed):
I'm still in disbelief that you tackled all of this. Looks great!

- I'm personally not a fan of snapshot testing, but with this particular library it might make a lot of sense because the markup changing constantly is hard to manage
- I might suggest making a simple npm script like `npm run update-snapshots` to regenerate them and add documentation to the contributing guide for it.
- I wouldn't let either of these things block merging, this is a fantastic update.

src/tabs/tabs.spec.tsx
jamesmfriedman: I updated this to return `new Promise` and resolve on pass

jamesmfriedman (approved):

PR #808, opened 2022-07-13
Fix textfield can't use floatLabel issue

hnsxxscyx:
adapter_ is undefined, should use adapter


develra:
It would be great to get this merged - at the moment this component crashes if you pass floatLabel in as a prop.

EmiBemi (approved):

PR #816, opened 2022-08-19
V14 updates for button

jamesmfriedman:
Starts the V14 update process and includes some quality of life features.

- Removes all generated files from git. `npm install` will now generate these on the fly. 
- Cleans any built files prior to running npm start. This will help avoid confusion with errors when trying to develop with built artifacts.
- Adds an .npmrc file to ensure we are pulling from the public registry.
- Adds node-sass, some of the newer material modules no longer have built dist css files
- adds the TouchTarget component
- Updates Button to V14


package.json
```diff
@@ -15,7 +15,7 @@
 	},
 	"scripts": {
 		"install": "lerna bootstrap --force-local --hoist",
-		"start": "export NODE_OPTIONS='--max-old-space-size=4096' && react-app-rewired start",
+		"start": "export NODE_OPTIONS='--max-old-space-size=4096' && npm run clean && concurrently \"react-app-rewired start\" \"chokidar './src/**/readme.tsx' -c 'npm run docgen:examples'\"",
```
jamesmfriedman, line 18: - Cleaning the directory before npm start so we don't run into the issue of our dev experiencing failing from built files
- Including concurrently and chokidar to watch readme files and recompile them. This means we get live updates of the docsite when running npm start. Note, this is only the examples, the prop tables won't update in development because it's an expensive operation to run.

src/button/index.tsx
```diff
@@ -32,6 +33,8 @@ export interface ButtonProps extends RMWC.WithRippleProps {
   danger?: boolean;
   /** Content specified as a label prop. */
   label?: React.ReactNode | any;
+  /** Makes the button more touch friendly. This will automatically be set true if used inside of TouchTargetWrapper.*/
+  touch?: boolean;
```
jamesmfriedman, line 37: Adding touch target support

src/button/index.tsx
```diff
@@ -91,7 +98,10 @@ export const Button: RMWC.ComponentType<
     return (
       <Tag tag="button" {...rest} ref={ref} className={className}>
         <ButtonRipple />
+        <ButtonTouch />
+        <ButtonFocusRing />
```
jamesmfriedman, line 102: The docs indicate that focus ring should always be present now.

.gitignore
```diff
@@ -35,3 +35,4 @@ yarn-error.log*
 
 report.*.json
 /rmwc
+generated-*
```
jamesmfriedman, line 38: Removing any generated files from the repo and rebuilding them as needed. This will avoid conflicts in the future and keep the codebase clean.

.github/workflows/version.yml
```diff
@@ -39,3 +39,5 @@ jobs:
       - name: 'Alpha Major Release'
         run: npm run alpha:major
         if: contains(github.head_ref, '/major') && github.base_ref == 'next'
+      - name: 'Build Docs':
+        run: npm run build:docs
```
jamesmfriedman, line 43: This was the missing script for updating the docs automatically on release

.npmrc
```diff
@@ -0,0 +1 @@
+registry=https://registry.npmjs.org
```
jamesmfriedman, line 1: This is necessary to prevent npm installs with private repos from overwriting package.lock

.github/workflows/version.yml
EmiBemi: Nice we god this fixed!

EmiBemi (approved):

PR #821, opened 2022-08-20
chore(fab): upgrade fab to v14

EmiBemi:


thaestrup (dismissed):

jensborch (approved):

src/fab/index.tsx
```diff
@@ -23,15 +23,8 @@ export interface FabProps extends RMWC.WithRippleProps {
...
-    const {
-      children,
-      label,
-      icon,
-      trailingIcon,
-      mini,
-      exited,
-      ...rest
-    } = props;
+    const { children, label, icon, trailingIcon, mini, exited, ...rest } =
```
jamesmfriedman, line 26: Do you have prettier installed? I think if not we'll keep incurring formatting only changes. I'd have to check but we might be missing a prettier config file in the repo.

jamesmfriedman (approved):
Yay for easy!

src/fab/index.tsx
EmiBemi: I do have prettier and it should be enabled.. hmm

PR #838, opened 2022-12-07
This should fix the build on Windows 10

jensborch:


jensborch:
I expect this to also work on Mac and Linux, but it would be nice if someone could test this @thaestrup @EmiBemi

jensborch:
@EmiBemi @thaestrup Tested ok on Mac, so I think it should be merged

EmiBemi (approved):

PR #840, opened 2022-12-12
fix(menu): Correctly renderToPortal for menu

EmiBemi:
* Fixes https://github.com/rmwc/rmwc/issues/800
* Reverts https://github.com/rmwc/rmwc/pull/741 as it proves to have introduced a new bug
* Added test to verify that portal is not mounted twice (however, this test fails as of right now)

I hope to find a fix to the 3rd bullet before merging.


develra (approved):
Simple enough! LGTM but I'm just some rando.

develra:
@jamesmfriedman -  would love to be able to get this merged and released in the next couple of weeks if you could take a gander. Thanks!

kintz09:
Did this fix happen to make it into release 8.0.4? Based on the changelog I don't believe it did. But if it did, I'd be happy to confirm the fix in my project.

PR #841, opened 2022-12-14
Fix tests for fixes to render menu to portal

develra:


src/base/index.spec.tsx
```diff
@@ -252,20 +252,17 @@ describe('Portal', () => {
         <>
           <Portal />
           <Button onClick={() => setOpen(true)}>Open</Button>
-          {open && (
```
develra, line 255: I've generally found conditionally rendering dialogs in RMWC to be pretty flaky on React 18 - in general I just let the control state decide if it should be visible or not, but happy to discuss!

src/base/portal.tsx
```diff
@@ -49,6 +49,5 @@ export function PortalChild({
   if (portalEl) {
     return ReactDOM.createPortal(children, portalEl);
   }
-
-  return <>{children}</>;
+  return null;
```
develra, line 52: Largest change is here - what should happen if portalEl is undefiend in PortalChild? I think just not rendering anything makes sense, but happy to discuss.

EmiBemi:
I tried running the docs locally with the changes here. It fixes the problem with the portal rendering twice, but not the problem with the menu positioning... :(

EmiBemi:
Okay, I managed to fix it combing some of the changes from this PR. Will merge and fix 👍

EmiBemi:
Nvm.. I was running the wrong test. So we still do not have a fix, sadly.

PR #842, opened 2022-12-20
Fix menu render to portal, ensure dialog only renders once in portals.

develra:
Ok! That ended up being more difficult than I had anticipated, but this solution.

1. Fixes the menu render in portals issue :partying_face: 
2. Doesn't break existing menu's or dialogs
3. Makes sure dialogs in portals only render once
4. Doesn't break backwards compatibility

It isn't the cleanest thing in the world, but not too bad either - very open to any feedback. 

I verified all looks good in docs and tests still pass- I enabled the 'render twice' test as well.


src/base/index.spec.tsx
```diff
@@ -236,10 +236,11 @@ describe('Portal', () => {
     const { asFragment } = render(<Portal />);
     expect(asFragment()).toMatchSnapshot();
   });
-  it.skip('does not mount twice', async () => {
+  it('does not mount twice', async () => {
     const Content = ({ value, inc }: { value: number; inc: () => void }) => {
       React.useEffect(() => {
         inc();
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
develra, line 243: this will infinite inc if you add the dep.

PR #852, opened 2023-02-19
fix(menu): Disable events when disabled is true

EmiBemi:
Fixes https://github.com/rmwc/rmwc/issues/847 by setting css pointer events to none.

### DIFF
--- a/src/list/list-item.css
+++ b/src/list/list-item.css
@@ -1,3 +1,8 @@
+/* MDC gap fill: according to MDC documentation, disabled items should have
+  0.38 opacity, but the style is not included, so adding a gap fill here. 
+  https://m2.material.io/design/interaction/states.html#disabled */
+
 .rmwc-list-item__disabled {
   pointer-events: none;
+  opacity: 0.38;
 }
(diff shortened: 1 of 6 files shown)


b-d-m-p:
@EmiBemi @jensborch So, I have confirmed that menu is functioning as it should, but the list item does not appear disabled. 
image

According to MDC documentation, disabled items should have `0.38` opacity, https://m2.material.io/design/interaction/states.html#disabled 
image


 but the style is not included, so I added an opacity to the RMWC class.  

```
/* MDC gap fill: according to MDC documentation, disabled items should have
  0.38 opacity, but the style is not included, so adding a gap fill here. 
  https://m2.material.io/design/interaction/states.html#disabled */

.rmwc-list-item__disabled {
  pointer-events: none;
  opacity: 0.38;
}

```

image


How should we handle the merging in this case? If @EmiBemi or @jensborch can confirm that my additions are alright, then I will merge the branch.

b-d-m-p:
Got feedback in Discord. Merging. 
image

b-d-m-p (approved):
Looks good to me. Merging.

Issue #853, opened 2023-02-28
Add support for RMWC Portal inside a Shadow Dom


kintz09:
This is a request for a feature improvement to the Portal component. I'd like the Portal component to be supported in web applications that live in a Shadow Dom.

## Use Case
I am the developer of a web application which offers a plugin architecture to render children apps within the parent application. The parent application and the children applications are all built using react. I am using the RMWC component library in both the parent application and the children applications. 

The parent app creates a Shadow DOM element and loads the child application and its stylesheets into that Shadow Dom encapsulating the child app's stylesheet from the parent app and other children apps.

In both the parent and children applications, there are times when I am choosing to render elements such as a Dialog or Menu to the portal. Each app using RMWC places its own instance of the Portal component in the appropriate location with the app's DOM.

## Current Problem
When one of the children applications sends a component to the Portal, it does not get placed into the correct Portal instance inside that application's DOM. Instead anything rendered to the portal will render into the Parent application's Portal instance. Initial debugging has me believing this occurs because the RMWC Portal component uses a [hardcoded element id](https://github.com/rmwc/rmwc/blob/04db5896e70155905f2019fa54c5f8e2e80f9a40/src/base/portal.tsx#L4).

- I add the `renderToPortal` prop on the `Menu` component which will give the prop a value of `true`.
- In the PortalChild's `useEffect`, the `renderTo` prop will have a value of `true` and get into the [logic at line 28](https://github.com/rmwc/rmwc/blob/04db5896e70155905f2019fa54c5f8e2e80f9a40/src/base/portal.tsx#L28).
- PortalChild will then grab the portal element by element id where the id is hardcoded to a value of `rmwcPortal`.
- This sets the `element` variable and bring us to line 45 where react state variable `portalEl` gets set.
- Now that `renderTo` and `portalEl` are both defined, PortalChild will create a portal to `portalEl` and render the menu into the Dom.

`document.getElementById` will return the first element it finds with the provided Id. The parent application's portal instance is the first in the DOM order, so this is the portal element returned and used even though the child application is the one trying to use its portal. Although, as I'll discuss later in this post, its not possible to get the child app's Portal element using `document`.

[...]

PR #855, opened 2023-03-03
feat(portal): Support Portal in a ShadowDom by using PortalContext

kintz09:
This PR is intended to solve #853.

Solution:

- Created a PortalContext to store the portal element and function to set element
- If PortalContext is not found, the Portal should work exactly the same as before
- Portal element will set `portalElement` in PortalProvider if PortalContext is found
- PortalChild will render to the `portalElementFromContext` if found and `renderTo` is `true`
- I added `PortalProvider` to `App` to demonstrate its capability
- I didn't add any tests, but all tests are still passing

Notes:
I believe this is my first PR for RMWC so I'm not expecting this to be approved, but am hoping it will be enough for the team to determine if this is an acceptable solution.

Since I was modifying an existing component, I didn't know if any other files were necessary. Please advise and I will update.

### DIFF
--- a/src/base/PortalContext.tsx
+++ b/src/base/PortalContext.tsx
@@ -1,4 +1,4 @@
-import { createContext, FC, ReactNode, useState } from 'react';
+import React, { createContext, ReactNode, useState } from 'react';
 
 export interface PortalContextData {
   portalElement: HTMLDivElement | null;
@@ -14,7 +14,7 @@ export const PortalContext = createContext<PortalContextData>(
   portalContextDefaultValues
 );
 
-export const PortalProvider: FC<ReactNode> = ({ children }) => {
+export const PortalProvider: React.FC<ReactNode> = ({ children }) => {
   const [element, setElement] = useState<HTMLDivElement | null>(null);
 
   return (
--- a/src/base/PortalContext.tsx
+++ b/src/base/PortalContext.tsx
@@ -1,0 +1,30 @@
+import { createContext, FC, ReactNode, useState } from 'react';
+
+export interface PortalContextData {
+  portalElement: HTMLDivElement | null;
+  setPortalElement: ((element: HTMLDivElement) => void) | null;
+}
+
+export const portalContextDefaultValues = {
+  portalElement: null,
+  setPortalElement: null
+};
+
+export const PortalContext = createContext<PortalContextData>(
+  portalContextDefaultValues
+);
+
+export const PortalProvider: FC<ReactNode> = ({ children }) => {
+  const [element, setElement] = useState<HTMLDivElement | null>(null);
+
+  return (
+    <PortalContext.Provider
+      value={{
+        portalElement: element,
+        setPortalElement: setElement
+      }}
+    >
+      {children}
+    </PortalContext.Provider>
+  );
+};
--- a/src/base/portal.tsx
+++ b/src/base/portal.tsx
@@ -1,51 +1,76 @@
-import React, { useRef, useEffect, useState } from 'react';
-import ReactDOM from 'react-dom';
+import {
+  useEffect,
+  useState,
+  useContext,
+  useCallback,
+  ReactNode,
+  MutableRefObject
+} from 'react';
+import { createPortal } from 'react-dom';
+import { PortalContext } from './PortalContext';
 
 const PORTAL_ID = 'rmwcPortal';
 
 export type PortalPropT = Element | string | boolean | undefined | null;
 
-export function Portal() {
-  const el = useRef<HTMLDivElement | null>(null);
+export const Portal = (): JSX.Element => {
+  const portalContext = useContext(PortalContext);
+  const setPortalElement = portalContext?.setPortalElement;
 
-  return <div ref={el} id={PORTAL_ID} />;
-}
+  const portalRef = useCallback(
+    (node: HTMLDivElement) => {
+      if (node !== null && setPortalElement) {
+        setPortalElement(node);
+      }
+    },
+    [setPortalElement]
+  );
+
+  return <div ref={portalRef} id={PORTAL_ID} />;
+};
 
 export function PortalChild({
   children,
   renderTo,
   menuSurfaceDomPositionRef
 }: {
-  children: React.ReactNode;
+  children: ReactNode;
   renderTo?: PortalPropT;
-  menuSurfaceDomPositionRef?: React.MutableRefObject<HTMLDivElement | null>;
+  menuSurfaceDomPositionRef?: MutableRefObject<HTMLDivElement | null>;
 }) {
   const [portalEl, setPortalEl] = useState<Element | undefined>();
+  const { portalElement: portalElementFromContext } = useContext(PortalContext);
 
   useEffect(() => {
     let element: Element | undefined = undefined;
 
-    if (renderTo === true) {
+    if (renderTo === true && portalElementFromContext) {
+      element = portalElementFromContext;
+    } else if (renderTo === true) {
       element = document?.getElementById(PORTAL_ID) ?? undefined;
 
-      !element &&
+      if (!element) {
         console.warn(
           'No default Portal found. Did you forget to include it in the root of your app? `import { Portal } from "@rmwc/base";`'
         );
+      }
     } else if (typeof renderTo === 'string') {
       element = document?.querySelector(renderTo) ?? undefined;
 
-      !element &&
+      if (!element) {
         console.warn(
           `The selector you provided for renderToPortal "${renderTo}" didn't find any elements.`
         );
+      }
     } else if (renderTo instanceof Element) {
       element = renderTo;
     }
+
     if (element !== portalEl) {
       setPortalEl(element);
     }
-  }, [renderTo, portalEl]);
+  }, [renderTo, portalEl, portalElementFromContext]);
+
   // if renderTo defined, render children if we have the portalEl, else don't render anything.
   // menuSurfaceDomPositionRef is used to position the menu at the correct location on the menuSurfaceAnchor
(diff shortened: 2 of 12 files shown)


kintz09:
Please let me know if there's any additional work I can complete for this feature request. Thanks!

src/base/portal.tsx
```diff
@@ -18,42 +35,49 @@ export function PortalChild({
...
+  menuSurfaceDomPositionRef?: MutableRefObject<HTMLDivElement | null>;
 }) {
   const [portalEl, setPortalEl] = useState<Element | undefined>();
+  const { portalElement: portalElementFromContext } = useContext(PortalContext);
 
   useEffect(() => {
     let element: Element | undefined = undefined;
 
-    if (renderTo === true) {
+    if (renderTo === true && portalElementFromContext) {
```
EmiBemi, line 46: Do you think there is a way to test the portal from context in `index.spec.tsx`? Would be great to verify that it works in scenarios with a portal context element and without.

src/base/PortalContext.tsx
```diff
@@ -0,0 +1,30 @@
...
+  portalElement: HTMLDivElement | null;
+  setPortalElement: ((element: HTMLDivElement) => void) | null;
+}
+
+export const portalContextDefaultValues = {
+  portalElement: null,
+  setPortalElement: null
+};
+
+export const PortalContext = createContext<PortalContextData>(
```
EmiBemi, line 13: From what I understand, it's up to the user of RMWC to use this context if needed. Do you think you'd be able to add some docs for this?

EmiBemi (commented):
Looks good to me tbh. Let me know what you think of my comments and if you need any help! Let's get this merged soon

src/rmwc/docs/views/app/index.tsx
```diff
@@ -175,38 +176,40 @@ export function App() {
...
-          <ThemePicker
-            selectedThemeName={theme}
-            onThemeClick={(themeName) => {
-              window.localStorage.setItem('rmwcTheme', themeName);
-              setTheme(themeName);
-            }}
-          />
-        )}
-      </AppBar>
+      <PortalProvider>
```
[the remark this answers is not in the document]
kintz09: PortalProvider is not required for the portal to work. I added it here so when you run project locally you can see it in use. While you won't see any change visually, if you use the React Developer Tools you can see PortalProvider in the virtual Dom and see the portal's context hook has values in instances of PortalChild.

Let me know if I should remove PortalProvider from this script.

src/base/PortalContext.tsx
kintz09: Yes, that is correct. I'd be happy to put some information together for the docs. Where at would you recommended this be added? A new page?

src/base/PortalContext.tsx
```diff
@@ -0,0 +1,30 @@
...
+  portalContextDefaultValues
+);
+
+export const PortalProvider: React.FC<ReactNode> = ({ children }) => {
```
kintz09, line 17: @EmiBemi Is this file okay or would it be preferred if everything in PortalContext.tsx was moved into portal.tsx?

src/base/portal.tsx
kintz09: I'm still new to writing tests, but let me see what I can put together! 😄

src/base/PortalContext.tsx
EmiBemi: That would be ideal I think. If you need help with that, you can write some docs and send it to me and I can place it on the site.

src/base/portal.tsx
EmiBemi: Try your best. Feel free to contact me on Discord if you need some help with writing tests!

src/base/PortalContext.tsx
EmiBemi: It's fine with me!

src/base/portal.tsx
kintz09: I added a test for each combination of using PortalProvider and renderToPortal (4 tests). Hopefully these tests are what we're looking for.

I couldn't figure out how to get a reference of the Portal element without somehow modifying the Portal element, so I went ahead and defined an interface named `PortalProps` so that I could pass in a data-attribute. You'll see that as a new change `portal.tsx`.

I was expecting to have to trigger the dialog to open in order for it to be rendered in the portal, but the dialog's `open` property had no effect on when the element was rendered in the DOM, just whether it was shown. I assumed this is expected and removed the user event triggering the dialog to open.

kintz09:
@EmiBemi I'm still a bit lost as to where and how I should be putting documentation together for this contribution. Could I take you up on the offer to finish that step for me? I hope the following information and partial documentation will be enough for you to create the final documentation and get this PR merged and released. Let me know if you have any questions for me.

# Problem and Solution Overview

## Problem
I utilize the RMWC library in a react application that get rendered into a Shadow DOM created by a different web application. I discovered that the Portal component doesn't work inside of a Shadow DOM due to the fact that PortalChild calls `document.getElementById` to retrieve the Portal element before rendering content into the Portal. A Shadow DOM isolates styles and elements from the rest of the DOM, so when PortalChild calls `document.getElementById("rmwcPortal")` it returns undefined and the element will not be rendered into the portal as expected. The element can't access the document to retrieve the element by id in this way due to the isolation provided by a shadow dom. If inside a shadow dom, the code would need a reference to the shadowRoot and then it could call `shadowRoot.getElementById("rmwcPortal")` to get the Portal element, but it will not find the "rmwcPortal" element if using `document.getElementById("rmwcPortal").

## Solution
To solve this problem, I created the PortalContext script. I use this context as a global location to store a reference to the Portal element along with a function to set the portalElement. I then refactored the Portal component to implement the PortalContext. The Portal component will use context to set the portal element using a callback in the ref prop of the Portal's root div element. Now there is a direct reference to the Portal element in PortalContext which can be used instead of retrieving the element with `document.getElementById("rmwcPortal")`.

Next I implemented portal context in the PortalChild component. I refactored the component to use the portal element from PortalContext if it is available. This eliminates the need to call `document.getElementById` which doesn't work inside the shadow Dom. If portalElement is null, then we fallback to the original code and retrieve the element by id from document. This implementation makes the PortalProvider optional to users of the RMWC library. 

If they have chosen to include PortalProvider in their DOM tree, then the PortalChild component will use the direct reference to the element instead of getting the element by id from the document. If PortalProvider is not found in the DOM tree, then portalElement will be null and the PortalChild component will retrieve a reference to the Portal element using `document.getElementById("rmwcPortal")`. 

This solves my issue with the Portal component inside of a Shadow Dom but makes usage of the PortalContext optional. Also, it won't break existing functionality for any users who simply upgrade without reviewing the changes.


# Documentation Template

## PortalProvider

### Description
The PortalProvider component is an optional component that provides a global context for the Portal element in the RMWC library. This context is used by the Portal component to retrieve the Portal element without relying on document.getElementById, which doesn't work inside a Shadow DOM.

### Usage
The PortalProvider component is used by importing it into your project and wrapping it around your application or component tree similar to TypographyProvider.

```javascript
import { PortalProvider, Portal } from 'rmwc';

function App() {
  return (
    <PortalProvider>
      <div>
        {/* Other components here */}
        <Portal />
      </div>
    </PortalProvider>
  );
}
```

In this example, the PortalProvider component is used to wrap the Portal component. Elements rendered to the portal will now render directly to the portalElement stored in the portal context instead of retrieving the Portal element using `document.getElementById()`.

The PortalProvider component is optional for most users. If you are using the RMWC library in a Shadow DOM and the Portal component is not rendering inside the correct Portal element, then you can use the PortalProvider component to ensure that the Portal element is retrieved using the PortalContext.

### Notes
- The PortalProvider component does not accept any props.
- The Portal component only needs to be a descendent of PortalProvider for it to be functional.
- Unlike ThemeProvider or TypographyProvider, the  PortalProvider component should only be used once in your application or component tree. (If you include multiple instances of PortalProvider, the context will only grab the context values from the first PortalProvider parent and the solution will break.)

# Other documentation needs
The only other aspect I could think of that may need documentation updated is the Portal props. With this PR I've added standard html props to the Portal Element (omitting id and ref so they are never overwritten). This allowed me to pass props into the Portal component in the jest tests. It will also improve the accessibility of the component by allowing the user user to set standard HTML element props required to be compliant with WCAG standards.

src/base/portal.tsx
b-d-m-p: @EmiBemi can you take a look at this and see if the tests are alright?

b-d-m-p:
@kintz09 I think it should be the last page in the menu after Portal. Also, we need to add explanation and reference to the PortalProvider documentation in the Menu and Dialogs documentation under the portal sections. Something like, "if you wish to use multiple portals or portals with shadow doms, see the PortalProvider documentation." Can you add those additions? Cheers

b-d-m-p:
@EmiBemi besides the documentation part, are the other areas ready for merging?

EmiBemi:
@b-d-m-p I promised @kintz09 to handle the documentation. I just need to get around doing it. I have a easter holiday starting from today, so I will have plenty of time. Expect this to be done soon!

EmiBemi:
Just added the docs. If you check into this branch and run `npm start` you should be able to see the docs on the site. Please verify that it works 🙏

kintz09:
@EmiBemi Looks good locally on my side. Thanks again for helping me finish this contribution!

EmiBemi (approved):

EmiBemi:
@kintz09 Gosh, we've been so busy with v14 and other things at work that we've missed a release of these changes. I will do this asap.

EmiBemi:
@kintz09 This has finally been released in 8.0.8

PR #861, opened 2023-03-23
Dependabot automerge

jensborch:
This should automatically merge minor and patch updated from Dependabot. It is nearly impossible to test Github Action changes locally, so there is a chance that this will break our current build. I have added Husky and a pre-commit hook to validate the Github Action files. We could add additional stuff to this like npm test and/or prettier.


EmiBemi (approved):
to be honest I do not know much about github action scripts, but I read the through the code and found nothing that seemed off to me.

b-d-m-p (commented):
Same same

b-d-m-p:
Oh, it was already merged. Duh

PR #977, opened 2023-07-10
Release/major 2023 07 10

EmiBemi:


b-d-m-p:
@EmiBemi @jensborch I'm a bit confused because this PR has v13 and v14 both in it. Were we planning on not releasing v13 and then just include it in v14 when we release that? I'm sure we discussed this, I just forgot, sorry.

jensborch:
> @EmiBemi @jensborch I'm a bit confused because this PR has v13 and v14 both in it. Were we planning on not releasing v13 and then just include it in v14 when we release that? I'm sure we discussed this, I just forgot, sorry.

I understand your confusion. We have manually bumped this to version 13 and the pipeline will then bump to version 14 upon release. We will thus not do a version 13 release.  Also, we have not worked on this for some weeks now due to summer vacation, but we are now starting up again. I will start with the pipeline so that we can get the tests running again.

jensborch (approved):

jensborch (approved):

PR #987, opened 2023-08-17
WIP: Fix/storybook v1 

b-d-m-p:
This doesn't work, but a lot of things are updated and a few issues fixed, so worth using as a starting point to get Storybook fixed.


src/drawer/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rmwc/drawer",
-  "version": "8.0.7",
+  "version": "8.0.8",
```
EmiBemi, line 3: Probably we don't want this on next, as we are doing a prerelease. Seems like my commit somehow made it into this PR

jensborch:
Fixed by https://github.com/rmwc/rmwc/pull/1341

PR #1008, opened 2023-08-23
Update README

EmiBemi:


jensborch (commented):
I think we should add https://contrib.rocks/preview?repo=rmwc%2Frmwc

jensborch (approved):

PR #1038, opened 2023-09-01
Readme fixes + little typescript update

EmiBemi:
Renaming the prop `content` to `overlay` is going to be a breaking change, but it is clashing with some other props + typescript is complaining. Let me know if you think it is an acceptable change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "@action-validator/cli": "^0.5.3",
         "@action-validator/core": "^0.5.3",
         "@babel/core": "^7.22.10",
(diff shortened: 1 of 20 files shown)


jensborch:
> Renaming the prop `content` to `overlay` is going to be a breaking change, but it is clashing with some other props + typescript is complaining. Let me know if you think it is an acceptable change.

I think this is an actable breaking change, using `overlay` seem to be consistent with terminology used elsewhere in RMWC

jensborch (dismissed):

jensborch (dismissed):

bjarkehs:
You are still referring to the prop as `content` in some of the documentation.

jensborch (approved):

PR #1040, opened 2023-09-11
feat(tooltip): Support of v14 tooltip

EmiBemi:


src/tooltip/foundation.tsx
```diff
@@ -19,59 +20,52 @@ export const useToolTipFoundation = (
...
           return rootEl.ref?.parentElement?.getBoundingClientRect() ?? null;
         },
         getAnchorAttribute: (attr) => {
-          return anchorEl.ref
-            ? (anchorEl.ref?.getAttribute(attr) as string | null)
-            : null;
-        },
-        setAnchorAttribute: (attr, value) => {
-          anchorEl?.setProp(attr as any, value);
+          return anchorEl.ref ? anchorEl.ref?.getAttribute(attr) : null;
```
jensborch, line 58: Can't you write this as 

```
anchorEl.ref?.getAttribute(attr) ?? null
```

as above

jensborch (changes requested):
One very small comment :-)

jensborch (approved):

src/tooltip/index.tsx
```diff
@@ -3,14 +3,28 @@ import * as RMWC from '@rmwc/types';
 import { classNames, createComponent, Tag } from '@rmwc/base';
 import { useProviderContext } from '@rmwc/provider';
 import { useToolTipFoundation } from './foundation';
-import { AnchorBoundaryType, CssClasses } from '@material/tooltip';
+import { AnchorBoundaryType } from '@material/tooltip';
 
 export type TooltipActivationT = 'hover' | 'click' | 'focus';
+export type TooltipAlignT =
```
bjarkehs, line 9: If you move this to a shared types file, then you can import it an use it in `constants.ts`.

src/tooltip/constants.ts
```diff
@@ -0,0 +1,15 @@
+export const ALIGN_MAP: {
```
bjarkehs, line 1: If you rewrite this as: 
```typescript
export const ALIGN_MAP: Record<TooltipAlignT,{ xPos?: number; yPos?: number; withCaretPos?: number }>
```

Then you cannot accidentally forget a key.

src/tooltip/index.tsx
```diff
@@ -23,12 +37,14 @@ export interface TooltipProps {
   leaveDelay?: number;
   /** Manually control the open state */
   open?: boolean;
-  /** Whether or not to show an arrow on the Tooltip. Defaults to `false`. */
-  showArrow?: boolean;
```
bjarkehs, line 27: Is this no longer possible to support or?

src/tooltip/readme.tsx
```diff
@@ -138,6 +156,26 @@ export default function Readme() {
         </>
       </DocsExample>
 
+      <DocsExample label="Alignment">
+        <>
+          <Tooltip overlay={'Align start'} align="start">
```
bjarkehs, line 161: [nit] The outer `{` is not necessary? So you could write these as `overlay="Align start"`

src/tooltip/index.tsx
EmiBemi: I can't really find any documentation for it.. but MCW do expose some methods for controlling a "caret". However, I think it is only for rich tooltips and there is no documentation (that I can find)

jensborch (approved):

PR #1074, opened 2023-10-09
Introduce nx

bjarkehs:
This PR introduces `nx` to manage the repository.

It also rewrites the docs site in Vite. All components are separate packages that are built using Vite.

There are some tasks that still need to be done, these can be found in `introduce-nx.md`. Stuff under "Porting to nx" have intentionally been left out for now. Stuff under "New features" is some of the benefits we get with `nx` that we should definitely implement. These two sections are intentionally left out in order to minimize the PR and keep it focused on a working monorepo that we can publish from.


package.json
```diff
@@ -1,110 +1,138 @@
...
-    "react-dom": "^18.2.0",
-    "react-ga": "^3.3.0",
-    "react-live": "^3.1.0",
-    "react-markdown": "^6.0.2",
-    "react-router-dom": "^6.3.0",
-    "react-scripts": "^5.0.1",
+    "ngx-deploy-npm": "^7.0.1",
+    "nx": "16.8.0",
+    "prettier": "^2.6.2",
     "sass": "^1.54.4",
```
EmiBemi, line 132: This is the dependency that is currently causing breaking changes, as we now have to install sass in order to use v14. Can we somehow check whether that is necessary anymore?

.github/workflows/unit-test.yml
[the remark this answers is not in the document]
jensborch: I will take a look, but I probably don't have time this weekend

jensborch:
Some snapshot seems to be in Jest format - e.g.:

rmwc\packages\base\src\lib\__snapshots__\index-ssr.spec.tsx.snap
rmwc\packages\base\src\lib\__snapshots__\index.spec.tsx.snap

bjarkehs:
Those were some old unused files. I have now removed them.

I'm a little confused why Sonar says there's no coverage information

jensborch:
Some of the tests are failing in React 18 https://github.com/rmwc/rmwc/actions/runs/6546397818/job/17776775648

jensborch:
Since the name of the test have change we need to update the branch protections rules on the next branch

jensborch:
This comment is going to be really annoying. With this PR we are loosing the git history of all the RMWC components. I don't think we can live with that. What do you think @EmiBemi and @b-d-m-p. `git mv` should have been used to move the various files and directories. The important part is that we retain git history for the components, I think we can live without proper history on all the other files.

jensborch:
> This comment is going to be really annoying. With this PR we are loosing the git history of all the RMWC components. I don't think we can live with that. What do you think @EmiBemi and @b-d-m-p. `git mv` should have been used to move the various files and directories. The important part is that we retain git history for the components, I think we can live without proper history on all the other files.

I have tried using `git mv`, but this doesn't make a difference. Git track files based on content so we can see the file history using `git log --follow`, but some tools doesn't use this like the timeline in VS Code. To conclude, we can't get around this unless we keep files in the same place, which isn't an option :-(

package.json
jensborch: Using Vite might fix this, we can verify after releasing a new alpha version

PR #1170, opened 2023-11-15
Added RC tooltip from version 8 for backwards combability

jensborch:


packages/rc-tooltip/.npmignore
```diff
@@ -0,0 +1,9 @@
+**/*.spec.*
+**/*.story.*
+**/*.tsx
+*.tsx
+readme.tsx
+generated-examples.json
+generated-props.json
```
EmiBemi, line 7: these files are actually not located here anymore, but maybe that's a general cleanup we need to do

packages/rc-tooltip/src/lib/__snapshots__/tooltip.spec.tsx.snap
```diff
@@ -0,0 +1,25 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
```
EmiBemi, line 1: old snapshot which should be deleted

packages/rc-tooltip/src/lib/__snapshots__/rc-tooltip.spec.tsx.snap
```diff
@@ -0,0 +1,25 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
```
EmiBemi, line 2: I'm still puzzled why the snapshots look like this. But I guess there is not much to do?

packages/rc-tooltip/src/lib/rc-tooltip.spec.tsx
```diff
@@ -0,0 +1,101 @@
...
+      </RCTooltip>
+    );
+  });
+
+  it('showArrow', () => {
+    const { asFragment } = render(
+      <RCTooltip content="tooltip" showArrow>
+        <span>test</span>
+      </RCTooltip>
+    );
```
EmiBemi, line 31: Aha! It's because we don't dover on the tooltip I'm guessing. So we need to hover before rendering the snapshots.

packages/rc-tooltip/vite.config.ts
```diff
@@ -0,0 +1,79 @@
...
+        'react',
+        'react-dom',
+        'react/jsx-runtime',
+        /@rmwc\/.*/,
+        /@material\/.*/
+      ],
+      output: {
+        outro: (chunk) => {
+          if (chunk.isEntry && chunk.name === 'styles') {
+            return "import './tooltip.css'";
```
EmiBemi, line 52: Should be rc-tooltip.css

packages/rc-tooltip/vite.config.ts
```diff
@@ -0,0 +1,79 @@
...
+          }
+          return '';
+        },
+        assetFileNames: (chunk) => {
+          if (chunk.name === 'styles.css') {
+            return 'tooltip.css';
```
EmiBemi, line 58: should be rc-tooltip.css

utils/readme/src/readmes/rc-tooltip.tsx
```diff
@@ -0,0 +1,204 @@
...
+  return (
+    <Docs
+      title="RC Tooltips"
+      lead="Tooltips display informative text when users hover over, focus on, or tap an element."
+      module="@rmwc/rc-tooltip"
+      styles={['@rmwc/rc-tooltip/tooltip.css']}
+      examples={examples}
+      addon
+    >
+      <DocsSubtitle>Basic Usage</DocsSubtitle>
```
EmiBemi, line 30: Should we add a section above here stating that this is not the official mdc tooltip?

EmiBemi:
Do we want to state in the general readme what's new?

EmiBemi (approved):

PR #1192, opened 2023-11-23
fix(docgen-markdown): updated script to work with Nx

jensborch:
Update Docgen markdown to work with Nx. Still need to copy generated files into the separate components, but take a look anyway @EmiBemi @b-d-m-p


EmiBemi (dismissed):

packages/avatar/README.md
```diff
@@ -0,0 +1,285 @@
...
+        -   __'@material/ripple/dist/mdc.ripple.css'__;
+
+Images
+
+```js
+
+<\>
+
+  <Avatar
+
```
EmiBemi, line 21: There is a lot of unnessecary spacing here. But I guess we can't do much about it?

EmiBemi (dismissed):

packages/avatar/README.md
jensborch: Yes, it is annoying with the spaces. I don't think we can fix it with the current markdown lib, but there might be other options. We could also add prettier to the build process or as a commit hock to fix this. This might be a good idea anyway, but might affect many files, so it should be in a separate PR.

jensborch: I have added prettier  to the script, but it doesn't fix all the issues

jensborch:
@EmiBemi This should be ready know. I didn't fine a nice way to handel the generate-markdown dependencies in nx.json, so it is done in the individual project.json files

EmiBemi (approved):

PR #1204, opened 2023-11-28
chore: write tests for theme utils

EmiBemi:


packages/theme/src/lib/utils.tsx
```diff
@@ -162,15 +162,15 @@ const hexToRgb = (hex: string) => {
       };
 };
 
-const luminance = (r: number, g: number, b: number) => {
-  var a = [r, g, b].map(function(v) {
+export const luminance = (r: number, g: number, b: number) => {
+  var a = [r, g, b].map(function (v) {
```
jensborch, line 166: Can't we use "let" her?

jensborch (changes requested):

packages/base/src/lib/utils/strings.tsx
```diff
@@ -1,6 +1,8 @@
 export const toCamel = (str: string) =>
-  str.replace(/(-[a-z])/g, $1 => $1.toUpperCase().replace('-', ''));
+  str.replace(/(-[a-z])/g, ($1) => $1.toUpperCase().replace('-', ''));
 
 export const toDashCase = (str: string) => {
-  return str.replace(/([A-Z])/g, $1 => '-' + $1.toLowerCase());
+  return str
+    .replace(/(^[a-z]+|[A-Z][a-z]*|[0-9]+)/g, ($1) => '-' + $1.toLowerCase())
```
jensborch, line 6: We can use `\d` instead of `[0-9]`

jensborch (changes requested):

packages/base/src/lib/utils/strings.tsx
EmiBemi: That works too.

packages/theme/src/lib/utils.tsx
EmiBemi: Sure

PR #1235, opened 2023-12-18
fix(drawer): fixes drawer getting stuck in open state

jensborch:
@EmiBemi Not to happy about the "solution", but it seems to fix the issue. What do you think?

### DIFF
--- a/packages/drawer/src/lib/foundation.tsx
+++ b/packages/drawer/src/lib/foundation.tsx
@@ -18,7 +18,6 @@ const useDrawerFoundationFactory = (
 ) =>
   function useDrawerFoundation(props: DrawerProps & React.HTMLProps<any>) {
     const focusTrapRef = useRef<FocusTrap>();
-
     const { foundation, ...elements } = useFoundation({
       props,
       elements: {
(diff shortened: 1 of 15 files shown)


packages/drawer/src/lib/foundation.tsx
```diff
@@ -28,8 +27,10 @@ const useDrawerFoundationFactory = (
       foundation: ({ rootEl, emit, getProps }) => {
         let previousFocusEl: HTMLElement;
 
-        const f = new MDCConstructor({
-          addClass: (className: string) => rootEl.addClass(className),
+        return new MDCConstructor({
+          //This delayes adding the class until next animation frame, fixes issue with draw getting stuck in open state
+          addClass: (className: string) =>
+            window.requestAnimationFrame(() => rootEl.addClass(className)),
```
ehpnykredit, line 33: Does this work:
`addClass: (className: string) => rootEl.ref?.classList.add(className),`

jensborch: Yes it seems so, but I of course would need update all methods, e.g.:

`
         addClass: (className: string) => rootEl.ref?.classList.add(className),
         removeClass: (className: string) => rootEl.ref?.classList.remove(className),
         hasClass: (className: string) => rootEl.ref?.classList.contains(className) || false,
`

jensborch:
@EmiBemi This is probably the best solution I can come up with. When debugging, I can see that isOpening() and isClosing() is true in https://github.com/material-components/material-components-web/blob/d76666ad4c53a619946d70b3991ca35ae5d87d04/packages/mdc-drawer/dismissible/foundation.ts#L77. This only happens when clicking fast. This means that the close() method will return immediately without adding any class. Delaying adding classes using requestAnimationFrame fixes this, but I am afraid that the reason is that mdc-drawer--animate mdc-drawer--opening is added after the check.

jensborch:
@EmiBemi @b-d-m-p I think we should merge this "hack", and get it released. 

This buys us some time to find a better solution. 

I have additionally be researching some more, and it seems like the handleTransitionEnd event handler in foundation.tsx disappears from the rootEl, and  this is the reason behind the issue, but still have no idea why this happens :-(

EmiBemi (approved):

PR #1240, opened 2023-12-20
feat: add sass

b-d-m-p:
Fixes https://github.com/rmwc/rmwc/issues/1183


jensborch (approved):

jensborch:
We need to add sass when Vite was upgrade, so this PR i no longer relevant

PR #1298, opened 2024-01-24
chore: ran NX migration to latest

jensborch:


package.json
```diff
@@ -95,32 +95,32 @@
...
     "@types/node": "^20.11.4",
     "@types/rc-tooltip": "3.7.10",
-    "@types/react": "^18.2.38",
-    "@types/react-dom": "^18.2.18",
-    "@typescript-eslint/eslint-plugin": "^6.19.1",
-    "@typescript-eslint/parser": "^6.19.0",
-    "@vitejs/plugin-react": "~4.2.0",
-    "@vitest/coverage-v8": "^0.34.6",
-    "@vitest/ui": "^0.34.7",
+    "@types/react": "18.2.24",
```
EmiBemi, line 113: This was downgraded. Same for `@types/react-dom` and `eslint`. Is that intentional?

jensborch: Yes, NX is compatible with certain versions, and thus running a NX migration might downgrade. I think this is mostly an issue with the build toolchain (Vite etc.), so maybe we shouldn't downgrade React. We probably also need to exclude Vite from Dependabot. I will try to update the PR

EmiBemi (approved):

PR #1366, opened 2024-03-18
fix(icon-button): forward props only once

### LINKED ISSUE #1299: IconButton onClick fires twice
- What RMWC Version are you using [major.minor.patch]: 14.0.7

- Name your build system [Webpack, Rollup...]: Vite

- Describe the bug with as much detail as possible: When I click an IconButton the click event is fired two times as seen here, [https://codesandbox.io/p/devbox/gallant-keldysh-6wrkyl](https://codesandbox.io/p/devbox/gallant-keldysh-6wrkyl)
Check the chrome devtools using F12 for the console.log (not the integrated codesandbox "Preview Console" button as this dosen't happen there although the same thing happens in the regular browser)

- What happened, and what was supposed to happen: After some investigation, I came to the conclusion that there's an event attached to both the `<button>` and  `<i>` tags and they're both fired at the same time, is the extra event on the `<i>` intended?

p-98:
# Motivation
Fixes #1299

# Problem description
When `onIcon` prop is not set, `IconButtonRoot` is rendered.
In IconButtonRoot, props were forwarded to `Tag` and `Icon` elments.
This led to event handlers being called multiple times.

# PR description
Forward props only to outer element, like in `IconButtonToggle`.

Adapted tests:
  - add test for props being forwarded only once
  - updated snapshots
  - snackbar foundation now detects icon-buttons properly


p-98:
This is my first pr, so feel free to give hint me if I missed something :)

EmiBemi (approved):

PR #1385, opened 2024-04-13
docs: update css import lists

p-98:
# Problem

I had broken styles after updating from an old version.

# Cause

Because I'm using next.js, I have to manually import the stylesheets listed in the documentation. These are outdated.

# Solution

I went through all rmwc packages by hand and assured that all files imported by `packages/<package>/src/styles.ts` are mentioned in the documentation.

I was a little unsure what to do with `list-collapsible`. The component itself is located within the `list` directory. There exists `collapsible-list.ts` and `styles.ts` containing style imports. However,
- `collapsible-list.ts` is a subset of `styles.ts`,
- `styles.ts` imports `collapsible-list.ts`,
- none of the three list docs pages mentions `collapsible-list.ts`.
I decided to reflect `styles.ts` as in the other list pages, but you should take a look at this one.

In a personal project I wasn't able to import '@rmwc/list/list-item.css from the list component, so this probably needs correction. (After further investigation it seems these styles end up in `styles.css`, I just don't know why).

In the snackbar component it reads:
```ts
import '@material/button/dist/mdc.button.css';
import '@material/ripple/dist/mdc.ripple.css';
```
while in all other components the following is used:
```ts
import '@rmwc/button/styles';
import '@rmwc/ripple/styles';
```
I decided to leave this change up to you. But if adapted, the docs need to be updated accordingly.

There is a docs page `touch-target.tsx` that I was unable to find in the web interface. I still adapted it to reflect the components `styles.ts`.

## Further discussion

Doing this manually took quite some time. Ideally, it would be possible to express the dependencies in a way that docs and code would be generated from it. This would prevent them becoming out of sync, although I don't know if/how this is possible.

Even if it can't be autogenerated, a notice in `CONTRIBUTING.md` might be added to draw the attention to the fact the docs must be updated when css imports are changed. I would even consider this type of change breaking(!) and mark it as such, since it requires code changes by the user.

### DIFF
--- a/utils/readme/src/readmes/avatar.tsx
+++ b/utils/readme/src/readmes/avatar.tsx
@@ -24,6 +24,7 @@ export default function Readme() {
       styles={[
         '@rmwc/avatar/avatar.css',
         '@rmwc/icon/icon.css',
+        '@rmwc/ripple/ripple.css',
         '@material/ripple/dist/mdc.ripple.css'
       ]}
       examples={examples}
--- a/utils/readme/src/readmes/button.tsx
+++ b/utils/readme/src/readmes/button.tsx
@@ -16,7 +16,8 @@ export default function Readme() {
       module="@rmwc/button"
       styles={[
         '@material/button/dist/mdc.button.css',
-        '@rmwc/@rmwc/icon/icon.css',
+        '@rmwc/icon/icon.css',
+        '@rmwc/ripple/ripple.css',
         '@material/ripple/dist/mdc.ripple.css'
       ]}
       docsLink="https://material.io/develop/web/components/buttons/"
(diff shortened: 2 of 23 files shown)


EmiBemi:
Thanks for the PR! I think for now we should just let collapsible-list.ts file be. As you said, it's part of the list styles import.

p-98:
Okay, then we'll leave collapsible-list.ts as is. On my side, two questions are left:

What about `list/list-item.css`, can you confirm I it ends up as `@rmwc/list/styles.css` in the production build?

How about the notice in `CONTRIBUTING.md`? Under "Contributing to Existing Components", something like
```
- When changes to the style imports are made, the docs in "utils/readme/src/readmes" must be updated. This should be marked as breaking change!
```

PR #1387, opened 2024-04-15
feature(chip): reintroduce old chip

EmiBemi:


packages/chip/README.md
```diff
@@ -1,37 +1,79 @@
-# Chips
+# Evolution Chips
```
jensborch, line 1: Does not look right

packages/chip/src/lib/chip.story.tsx
jensborch, line 1: We should have a Storybook for the old chip (and the new)

jensborch (changes requested):

jencnykredit (approved):

jensborch (approved):

PR #1422, opened 2024-05-27
Migrated to NX 19

jensborch:
This should upgrade NX to version 19 and ESLint to new ESLint Stylistic configuration https://eslint.style/guide/getting-started. I am a little unsure about the ESLint changes, so review them thoroughly :-)


.eslintrc.json
```diff
@@ -24,12 +24,16 @@
     {
       "files": ["*.ts", "*.tsx"],
       "extends": ["plugin:@nx/typescript"],
-      "rules": {}
+      "rules": {
+        "@stylistic/no-extra-semi": "error"
```
EmiBemi, line 28: Was this rule something we had before?

EmiBemi (approved):

.eslintrc.json
jensborch: My understanding is that this rule was inherited from NX, but with stylistic this is no longer the case and we should thus added it, but I am not 100% sure

PR #1486, opened 2024-07-02
Fix NX release

jensborch:


EmiBemi (approved):
Does it go well with a dry run?

jensborch:
@EmiBemi You can pass dry-run as described her: https://nx.dev/nx-api/nx/executors/run-commands#chaining-commands-interpolating-args-and-setting-the-cwd. This will not work for the "nx release changelog", but I hope we can live with that

EmiBemi (approved):

PR #1572, opened 2024-10-21
fix(select): translate value

EmiBemi:


packages/select/src/lib/getTranslatedValue.ts
```diff
@@ -0,0 +1,36 @@
+import { FormattedOption, OptionsType } from './select';
+
+export const getTranslatedValue = (
```
jensborch, line 3: Maybe we should call this getDisplayValue or getFormattedValue

jensborch (changes requested):

packages/select/src/lib/select/foundation.tsx
```diff
@@ -417,7 +426,11 @@ export const useSelectFoundation = (
   // Default value
   useEffect(() => {
     if (defaultValue) {
-      setSelectedTextContent(defaultValue.toString());
+      const translatedDefaultValue = getDisplayValue(
```
jensborch, line 429: The const should be named defaultDisplayValue

jensborch (approved):

